PR #114, opened 2016-01-20
Car% added to pct-Shiny lines & zones popup

audev:
Sorry for the quick copy-paste; it's confused GH.

The real changes are in lines 102-104 (car% placeholder for lines) and 214-219 (placeholder and values for the centroids popups; this one must be pulled from the rebuilt c.Rds files).


Robinlovelace:
Warning: don't push to production

PR #115, opened 2016-01-21
Car% added to route network too

audev:
@Robinlovelace Tested on Liverpool, showing fine for:  Lines | Centroids  |  Route network (f&q), no crashes.

I didn't commit or close issue yet, I will keep an eye on it. 

Thanks @usr110

### DIFF
--- a/master/pct-shiny-funs.R
+++ b/master/pct-shiny-funs.R
@@ -155,7 +155,7 @@ routePopup <- function(data, scenario){
                     <td>Route Type</td>
                     <td>%s</td>
                     </tr>'),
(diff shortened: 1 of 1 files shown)


Robinlovelace:
Great work @audev . I've just tested and it works on the server:
http://geo8.webarch.net/liverpool-city-region/

One thing: please can you update it so it says '% who drive' instead of the current text 'Car users (perc)' in the centroid popup and the lines popup (where there is also a typo - it currently says 'Car users(perc)').

audev:
Added on fork. You can commit directly to master or I'll do early tomorrow.

Robinlovelace:
Many thanks

PR #120, opened 2016-01-30
Folder restructure

nikolai-b:
This moves all the starting regions into a `regions_www` folder and move the master files into the base (in `ui-base` and `server-base`).

Shiny server currently is set to serve the whole of the pct-folder which seems a security risk, this change means it would only serve the `regions_www` folder.  It also separates the regions from the rest of the code making the repo cleaner.

This cannot be deploy without changing the server!
The `shiny-server.conf` config needs updating to:
```
  site_dir /pathTo/pct-shiny/public_www;
```
and the cronjob needs updating (and testing)

Also this will require changes in `pct-load`


Robinlovelace:
Good stuff - will pull once pct-load is updated.

PR #124, opened 2016-02-01
Popup update

nikolai-b:
Moved into a PR


Robinlovelace:
Sweet - thanks for plonking into a PR @nikolai-b - will follow your lead for future commits. Assumed these minor addition were too small to merit a PR though, non? Thoughts @usr110 and @audev? Relates to conversation on Friday and quite important in relation to how we proceed, this question. Great that the updates eventually got included! Thanks all.

PR #131, opened 2016-02-07
Rename cols

nikolai-b:
This updates shiny inline with the changes in https://github.com/npct/pct-load/pull/34
Suffolk is using the new column names so this change should work there (but no where else)
If you get time can you test it for Suffolk and then merge this?

P.S. @Robinlovelace I've removed you folder move from this branch as it isn't a rename and I think it was included by accident (sorry, you might have to drop your local branch to pull this down...)

### DIFF
--- a/pct-shiny-funs.R
+++ b/pct-shiny-funs.R
@@ -38,7 +38,7 @@ lineColNames <- c(
 
 numericZoneColNames <- c(
   "Average slope"                     = "avslope",
-  "Cyclists in census"                = "base_olc",
+  "Cyclists in census"                = "Bicycle",
   "Cyclists at govenment target"      = "base_slc",
   "Increase at govenment target"      = "base_sic",
   "Cyclists at gender equality"       = "gendereq_slc",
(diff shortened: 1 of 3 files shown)


Robinlovelace:
@nikolai-b great work, will test then merge now.

Robinlovelace:
@nikolai-b have you tested it? On my machine in the rename-cols branch I get 

```
Warning in min(values) : no non-missing arguments to min; returning Inf
Warning in min(x, na.rm = na.rm) :
  no non-missing arguments to min; returning Inf
Warning in max(x, na.rm = na.rm) :
  no non-missing arguments to max; returning -Inf
Warning: Error in round: non-numeric argument to mathematical function
    72: sprintf
...
```

Robinlovelace:
Apologies Nikolai - I just didn't pull from the new branch before testing. When I did, it words :8ball:

Robinlovelace:
@nikolai-b pushed to production, seems to be working.

nikolai-b:
Have you rebuilt all the regions? I was finding a few crash.

Robinlovelace:
Yes I thought I did. Which crash?

Robinlovelace:
- we should have an automated test for this - should be quite easy with runApp()

PR #132, opened 2016-02-09
Update regions

nikolai-b:
We are not using Local Authorities (LA) now but regions.
This allows the region to change when the map moves.
It also deals with re-drawing the map when it moves far more elegantly using `reactiveValues` which means when the region is changed this reactive value triggers the observes to re-render (and the model tab to update).


Robinlovelace:
Great stuff - just tested it and it works. Ready to merge. Only thing: it wasn't changing regions on my local computer when I scrolled outside the region. Any ideas why? Ah just figured it out: we need to delete all the `isolated` files from the region directories - suggest we need another solution as this one persists even when you change the value! Oops.

https://github.com/npct/pct-data/blob/master/berkshire/isolated

Robinlovelace:
That isolated issue was sorted here: https://github.com/npct/pct-data/commit/2c544d67b67b73f1290cb9ef3697dbb68b2cc673

Robinlovelace:
@nikolai-b still doesn't seem to be changing to new regions checkout out in update-regions. Am I missing something?

nikolai-b:
works for me.  There are a few remaining isolated files
`find -name 'isolated' -type f -delete`
wipes them, I guess you were on leeds https://github.com/npct/pct-data/tree/e4eb205/leeds

Robinlovelace:
@nikolai-b yes but those remaning isolated want to stay there I think - leeds, for example, is not a standard region but a custom/la geographic level.

nikolai-b:
@Robinlovelace sorry, fixed now

Robinlovelace:
Tested. Works. Is superfast. I like the fact that if we later decide not to use it it's just a case of added `isolated = TRUE` to stop that behaviour in the data. Will merge. Can you push to production and check it doesn't cause any unexpected side effects?

PR #137, opened 2016-02-18
Labels consistency and info shown per scenario

audev:
@Robinlovelace   Changes following James preferences on wording and info shown. Tested yesterday with latest Bedforshire data, no issues seen.

Still pretty sure still a bit more can be added re e.g. route types, so I won't close the issue yet.

### DIFF
--- a/pct-shiny-funs.R
+++ b/pct-shiny-funs.R
@@ -55,8 +55,8 @@ zoneColNames <- c(
 )
 
 scenariosNames <- c(
-  "olc"      = "Cyclists (observed)",
-  "slc"      = "Cyclists (scenario)"
+  "olc"      = "Cycle commuters (Census 2011)",
+  "slc"      = "Cycle commuters (scenario)"
 )
 
 # Normalise the data ready for plotting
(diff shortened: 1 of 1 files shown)


nikolai-b:
I was thinking it'd might be easier to keep track if we passed in named vectors into the table generating functions, e.g.
```R
tableCommon <- function(dataLables){
  paste0('
     <tr>
       <td> Total commutes </td>
       <td> ', dataLabels['totalCommutes'], '</td>
     </tr>
     <tr>
       <td> Cycle commutes </td>
       <td> ', dataLabels['cycleCommutesAbs'], '(', datadataLables['cycleCommutesPerc'], '&#37;) </td>
     </tr>
     <tr>
       <td> Cycle commutes (scenario) </td>
       <td> ', dataLabels['cycleCommutesAbs'], ' </td>
     </tr>
     <tr>
       <td> Increase in commutes &nbsp; </td>
       <td> ', dataLabels['cycleCommutesInc'], ' </td>
     </tr>
     ')
}
```

Which could be called with `tableCommon('totalCommutes' = 422, 'cycleCommutesAbs' = 21, ...)`  What do you think?  I like code reuse here as it means the tables always appear similar to users.

Robinlovelace:
I'd say this is definitely better than using sprintf. I'm not sure how the `dataLabels` object (which is not yet in our system) would be created but think your plan makes sense in the example code above. How to create `dataLabels` is the key question I believe. I would also be cool to use `kable`.

Reproducible example (try it):

```
knitr::kable(cars[1:2,], format = "html")
<table>
 <thead>
  <tr>
   <th style="text-align:right;"> speed </th>
   <th style="text-align:right;"> dist </th>
  </tr>
 </thead>
<tbody>
  <tr>
   <td style="text-align:right;"> 4 </td>
   <td style="text-align:right;"> 2 </td>
  </tr>
  <tr>
   <td style="text-align:right;"> 4 </td>
   <td style="text-align:right;"> 10 </td>
  </tr>
</tbody>
</table>
```

nikolai-b:
I is a parameter you'd pass into the function, I gave a calling example above

Robinlovelace:
Sounds reasonable - a working example will clarify things. I think the simplest way is to create a function that outputs the `data.frame` of the popup we need depending on the context (e.g. lines, routes) and then call that in `kable`. But happy to see another approach. Did you test the kable code? Think it will make updating the popups much easier.

audev:
@Robinlovelace  To clarify: so the cycle % you mention in #140  (and consistency issues) were included in the PR code.
So now it's just about adapting it to use kable rather than sprintf ?

Robinlovelace:
Ah sorry. Yes in that case yes it's just changing it kable in the PR. Let me know when it's ready to test and I'll merge it ASAP. Thanks.

PR #142, opened 2016-02-24
Make variable names concise and consistent

Robinlovelace:
Please test this @audev and liaise with @JDWoodcock: the updated names are silly long, clutter up space and duplicate information (we say it's commters in the 2011 census in the key and other places - no need to say it for every line).

Plus this update will make it easier to add new layers that are not based on the 2011 Census (e.g. Manchester) - it will make our code more portable.

Before this PR: 

selection_081

After:

selection_080

### DIFF
--- a/pct-shiny-funs.R
+++ b/pct-shiny-funs.R
@@ -55,8 +55,8 @@ zoneColNames <- c(
 )
 
 scenariosNames <- c(
-  "olc"      = "Cycle commuters (recorded)",
-  "slc"      = "Cycle commuters (scenario)"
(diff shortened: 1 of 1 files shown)


Robinlovelace:
@nikolai-b any views on this?

JDWoodcock:
I agree we can replace 'Cycle Commuters' with 'Cyclists' and 'Car Commuters' with Drivers, as it does say total 'Commuters' above.

Robinlovelace:
Great - I'll merge this PR later today, pending further comments from others. Great to be having this discussion. Thanks.

PR #144, opened 2016-02-26
Remove selectize

nikolai-b:
This causes the select inputs to act as text inputs.  
This means the site is much harder to use on small devices.


Robinlovelace:
Sounds good - seems ready to merge. Not 100% sure what the difference is - easier to use on smartphones? Cheers!

PR #145, opened 2016-02-26
Try to rebuild front page map

nikolai-b:
@Robinlovelace not sure if you have the front page map generation script....?


nikolai-b:
I didn't want to spend too much time on it if you had a working script.  The map looks ok but there is no legend.

PR #159, opened 2016-03-06
Do not forcefully checkout data

### LINKED ISSUE #156: Don't delete new data when running the app

Robinlovelace:
Note: this will only show a warning in the R
console and not in the UI. Closes #156


Robinlovelace:
Should have checked this before as it just caused the loss of more test data...

PR #165, opened 2016-03-09
Labels consistency & use of kable

audev:
@Robinlovelace  Latest changes on labels consistency +use of kable. 

Pls notice that the data on intrazonal flows is not shown, so some values will appear as NA for now.

### DIFF
--- a/pct-shiny-funs.R
+++ b/pct-shiny-funs.R
@@ -94,7 +94,7 @@ tableCommon <- knitr::kable(data.frame(
 
   Value =    c("%s", "%s (%s%%)" , "%s (%s%%)",
                 "%s", "%s", "%s",
-            " %s", "%s Tm")), format="html", col.names=NULL)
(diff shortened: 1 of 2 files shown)


Robinlovelace:
Great work Alvaro! I've tested it and it works. Only mini issue are the units: can `Value of deaths avoided (pounds/year)` be `Value of deaths avoided (£/yr)`? Also, what is Tm in `Change in CO2 (e/year):	 -8.2 Tm`? maybe you mean Tonnes? Also I would put the units before the number to be consistent. Finally, the distance seems to be indented differently (see below). But these are minor issues. Overall it's an impressive achievement implemented sensibly with kable. Great work - will look to merge asap.

selection_095

audev:
@Robinlovelace  Tonnes symbol modified. I also prefer £ to pounds, but we haven't found the escape character, it doesn't look great w/o it.

PR #168, opened 2016-03-20
Tidy labels

nikolai-b:
The £ sign causes issues, safer to use the unicode point (sadly).


Robinlovelace:
Have you tested this on production?

Issue #184, opened 2016-03-30
Add 'Transparent zones' radio button


Robinlovelace:
Just discussed with James and Rachel (happy for more discussion here). The opacity of the colour means it's hard to identify road names and other important features of the underlying map. Clicking it would make the zones see-through.

@usr110 would you be up for taking a look - suspect it could be a 2 liner.

RachelAldred:
Hi @Robinlovelace this came up again while going over London and I agree this would be a useful feature :)

usr110:
I had a brief talk with Nikolai about it. Like me, he too, doesn't seem very keen on adding an additional control. However, I am now fixing the issue. We can discuss more in the technical meeting.

PR #197, opened 2016-04-15
Add zone transparancy checkbox

nikolai-b:


nikolai-b:
Refs #184 
When the layers are selected the colour can obscure road names.
There are two solutions: 

* add an additional check box (this PR)  
* make the zones more transparent

Thoughts?

Robinlovelace:
@nikolai-b we've already done this and there's agreement that it's way better. If you'd like to change it on production please put in a pull request that implements one of the above options. I've already implemented the first one so I guess it's up to you to increase the transparency, test, and see if people like it.

PR #204, opened 2016-04-22
Fix returning to map tab when bounds are null

nikolai-b:
This was causing crashes when changing tabs
as the map bounds was momentarily null.


nikolai-b:
I was getting this error in the console:

```
Warning: Error in <Anonymous>: unable to find an inherited method for function ‘spTransform’ for signature ‘"NULL", "CRS"’
    66: <Anonymous>
    65: spTransform
    64: sortLines
    63: plotLines
    62: function_list[[k]]
    61: withVisible
    60: freduce
    59: _fseq
    58: eval
    57: eval
    56: withVisible
    55: %>%
    54: observerFunc
     1: runApp
```

PR #205, opened 2016-04-22
Remove transparent zones checkbox

nikolai-b:
@Robinlovelace I know this is controversial but I find the decreased zone opacity in this PR is a good balance between seeing areas cycling level and reading the road names.

@usr110 thoughts also?

### DIFF
--- a/server-base.R
+++ b/server-base.R
@@ -153,6 +153,7 @@ shinyServer(function(input, output, session){
 
       if (event$group == "centres"){
         leafletProxy("map") %>% addPolygons(data = toPlot$zones[toPlot$z$geo_code == id,],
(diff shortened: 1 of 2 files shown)


Robinlovelace:
Hi Nikolai, we need to discuss this before merging it. People seem to like the check-box. I'll send around an email about it now to get a temperature test.

nikolai-b:
I would like the user interface as simple and clean for people to use as possible.  When we all met up in Westminster I asked if we could not have a checkbox for transparency of the zones but instead pick sensible values ourselves and that was agreed then.  I'm sure things have changed but I'd really like to keep things simple and be very strict about simply adding more things to the interface.

PR #206, opened 2016-04-22
Remove the zone popups when any lines are shown

### LINKED ISSUE #191: Clicking on lines and getting areas instead
I've been finding it a bit of a struggle to manage to click on route network lines and have been getting the zone data instead. I realise with so many potentially clickable links it might be hard to improve this, so wondered if one option might be for when we're showing links, to turn off the area information by default (?) @Robinlovelace @JDWoodcock @nikolai-b

nikolai-b:
Fixes #191

### DIFF
--- a/server-base.R
+++ b/server-base.R
@@ -238,6 +238,8 @@ shinyServer(function(input, output, session){
   # This function updates the zones and the lines
   observe({
     region$current
+    showZonePopup <- (input$line_type == 'none')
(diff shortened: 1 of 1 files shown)


Robinlovelace:
Tested, works - NB: this should be documented in the [User Manual](https://github.com/npct/pct-shiny/blob/master/static/FAQs.html). You up for adding that Nikolai? Great work.

PR #213, opened 2016-04-26
Export all lines in geojson

### LINKED ISSUE #211: geojson export
Hi all, Jonathan from @GMcycling here.

The geojson export works fine as-is, but is there any way to export 100% of the data for an area, or is this already what's exported when viewing the entire area regardless of what "Percent (%) of Network" is set to?

I don't see the benefit of the export being limited by either the slider for N. Lines (most cycled) / Percent (%) of Network or the visible area. The former could lead to people working with an incomplete dataset, and while the latter may occasionally be useful I feel anyone dealing with an export is likely to be able to crop it themselves if required.

nikolai-b:
Closes #211 (NOTE: only exports quite lines when both are shown)

@Robinlovelace you might be able to get `rbind` to work but its just black magic to me, the ids clash, then the numbers of columns of arguments do not match then the row.names of data and Lines IDs don't match.  :confounded:

### DIFF
--- a/server-base.R
+++ b/server-base.R
@@ -459,8 +459,14 @@ shinyServer(function(input, output, session){
     # the argument 'file'.
     content = function(file) {
       # Bug in writeOGR that there can be no "." in the file name
+      output <- switch(input$line_type,
+                       'straight' = toPlot$l,
+                       'route'    = toPlot$rQuiet,
+                       'd_route'  = toPlot$rFast,
+                       'rnet'     = toPlot$rnet
+      )
       fileNoDot <- unlist(strsplit(file, ".", fixed = T))[1]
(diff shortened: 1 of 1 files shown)


Robinlovelace:
Fantastic work @nikolai-b. I can help combine those pesky fast routes: `tmap::sbind()` would do it or you could add the additional step of `sp::spChFIDs(rq) = x`. Let's do it Friday :8ball:

Robinlovelace:
Hey @nikolai-b I've looked into it also. I think it's fine like this but that we need to say 'Only exports quietest routes' somewhere. Thinking is 2fold:

1. We already allow rf to be downloaded, this way we avoid duplication
2. Currently rq has fewer columns than rf, making it tricky

PR #214, opened 2016-04-27
Update quietness diversion factor popup text

Robinlovelace:
This small change removes the % symbol from the right hand column, to make it consistent with (hopefully) all other popups. Tested.


nikolai-b:
:+1:

Issue #243, opened 2016-05-14
Add download csv functionality


Robinlovelace:
- [ ] For lines
- [ ] For areas

I envision this looking something like this in the lines data tab (text can be smaller):

![screenshot from 2016-05-14 17-47-32](https://cloud.githubusercontent.com/assets/1825120/15269613/5630eeb4-19fc-11e6-89d0-098090c8e9ad.png)

Robinlovelace:
Discussed with @nikolai-b who may be up for adding the sha to the download link in pct-shiny.

JDWoodcock:
Can we also have shape file option @RachelAldred

Robinlovelace:
That's already in there as geojson. @JDWoodcock geojson is the open source equivalent of a shapefile and is more portable: 

![screenshot from 2016-05-16 19-47-00](https://cloud.githubusercontent.com/assets/1825120/15300142/0584e41c-1b9f-11e6-9dd8-271138f95174.png)

I think the main issue is communicating *how* to download the data and what you can do with it. @RachelAldred and I are on it, as described here: https://github.com/npct/pct-shiny/issues/244

PR #247, opened 2016-05-16
Replace download handler with links to data

### LINKED ISSUE #246: Download data seems to be broken
Just tried for west-yorkshire and it send me [here](http://pct.bike/session/521ab8a2cd8c20b08df1bdfcef36267b/download/downloadData?w=), with the message:

```
Page not found

Sorry, but the page you requested doesn't exist.
```

Any ideas @nikolai-b ?

nikolai-b:
Remove the download option in the map, instead add links on the zone and line pages.  Should fix #246.


nikolai-b:
Also partially fixes #243 (but we need to regenerate all the geojson and push a static version to github).

usr110:
Fails on this [line](https://github.com/npct/pct-shiny/blob/e1217b5f31a4879e5191148db1336170e28360a7/server-base.R#L30) as no such file exists (on Windows at least). Does one always need to create it with the repo's SHA?

I may be out of the loop, but are we ditching line data to be downloaded as 'geojson' files?
https://github.com/npct/pct-shiny/blob/e1217b5f31a4879e5191148db1336170e28360a7/server-base.R#L379-L397

Robinlovelace:
Hi Ali, good question. The plan now is to download all files as static files that are pre-generated, .csvs (as described [here](https://github.com/npct/pct-shiny/issues/243)] and geojsons. Reasons:

1. People will almost always want all lines for an area
2. It's useful to have a static file to point people to outside the PCT
3. Less work for the server

The downside is that it will be more data stored in pct-data but we can live with that.

usr110:
Robin, thanks for the explanation. I did have a look at the [issue](https://github.com/npct/pct-shiny/issues/243) you mentioned. What I was asking is: why geojson files were _not_ offered to users to download any more? Unless I miss something, this pull request removes the code which allows geojson files to be downloaded.

nikolai-b:
@usr110 yes, sorry I didn't explain.
I need to add generating simplified geojson for Rnet, rf, rf, l and z (and a csv if it is missing) to pct-load.
Then I need to cycle though the regions and create these files, upload and upload them to the data repo.  Then add a download link here!
So the plan is to keep geojson download.
Does this fail still on windows?  Could you give the output if it fails so I can debug?
[This line](https://github.com/npct/pct-shiny/blob/e95be3a04f1147afab772903f0f01f32262ad50c/scripts/init.R#L19) should create the `repo_sha` file but I had things in the wrong order and just pushed a commit that might fix it.

Robinlovelace:
Agreed - @nikolai-b will add the new geojson data ready for download. Will be great if this can be done by the end of tomorrow. Thanks!

Robin

usr110:
Thanks @nikolai-b for the explanation. It's now clear. 
I might test the branch again on my computer to see if your fix worked or not.

Robinlovelace:
Errors for me. Please try to reproduce @nikolai-b - keen to pull this but not seeming to work atm for me on my laptop:

```
> runApp('regions_www/kent')

Listening on http://127.0.0.1:3350
[1] "Fetching data"
Warning in eval(expr, envir, enclos) :
  Checking out a the version of the data saved in the data sha
Warning: Error in paste: object 'data_sha' not found
    47: paste
    46: system2
    45: eval [../../scripts/init.R#17]
    44: eval
    43: withVisible
    42: source
    41: eval [../../server-base.R#35]
    40: eval
    39: withVisible
    38: source
     1: runApp
Error in paste(c(env, shQuote(command), args), collapse = " ") : 
  object 'data_sha' not found
```

usr110:
I guess it's to do with 'data_sha' being created _after_ 'init.R' is called where it's being used.

Robinlovelace:
@nikolai-b please test stuff before putting in PR's - assuming this wasn't tested.

Robinlovelace:
@nikolai-b please test stuff before putting in PR's - assuming this wasn't tested.

nikolai-b:
Yes, but must have had the `data_sha` in my local env somewhere.  @usr110 can you check it works on Windows.

usr110:
@nikolai-b Yes, it's now working. Thanks for the fix.

PR #249, opened 2016-05-19
Zone popup on mouse over event

usr110:
Allows users to display zone information on mouse over event. Aesthetics can be improved, but other than that, it is working fine.
Please test it out and share your feedback - especially you @nikolai-b

### DIFF
--- a/server-base.R
+++ b/server-base.R
@@ -186,7 +186,7 @@ shinyServer(function(input, output, session){
 
     leafletProxy("map")  %>% clearGroup(., "straight_line") %>%
       clearGroup(., "quieter_route") %>% clearGroup(., "faster_route") %>% clearGroup(., "route_network") %>%
-      removeShape(., "highlighted")
+      removeShape(., "highlighted") %>% removeShape(., "mouse-over")
 
     leafletProxy("map") %>% {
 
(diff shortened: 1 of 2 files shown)


nikolai-b:
Hey, it looks good!

Major issue:
How do we do this with lines and routes?  

We could leave it so that the info box remains but the centroids and lines are clickable, and then the info box displays that.  I much prefer the info box to a popup!
The problem is that this means interacting in different ways, one with a mouse over the other with a click but I don't think that's too complicated.  OR we could still use clicks but have that be the action to select a zone / line / centroid.


Very minor issue:
You can still click on zones.  Clicking on two different ones crashes the app
```
Warning: Error in if: argument is of length zero
    55: isolate
    54: observerFunc [../../server-base.R#129]
     1: shiny::runApp
```

usr110:
@nikolai-b 
Thanks for testing and sharing your comments. For now I have removed the click feature for zones - and also fixed the error ([here](https://github.com/npct/pct-shiny/pull/249/commits/7d9fc49f80fcee731060aa54609b76489d2f97dc)) . Please try it again.
As for making this as a wider replacement for zone popups, I agree that it's a good idea. However a combination of mouseover and mouseclick functionalities might make it a bit difficult for users. For a test, I enabled the mouseover functionality for when lines are selected. This made it difficult to select lines. For now I'd be happy to leave it as is. In any case, let's talk more about it in our meeting - possibly involving wider team.

Robinlovelace:
Great work @usr110 but I cannot merge this yet as it seems the outputs clash with the UI box:

selection_165

Robinlovelace:
I don't think we should prioritise this - please can you return to it once the layer switching is sorted - more important atm imo.

usr110:
Thanks for testing it out, Robin. Re priority, I agreee.

Robinlovelace:
@usr110 there is a new way to do this: http://rpubs.com/bhaskarvk/kml 

Closing this PR as I think it's gone stale and that this branch has diverged.

Should be easier to do now, with fewer lines of code, but please re-open this if you think this is the best way forward. Cheers.

PR #260, opened 2016-05-23
Order lines / flows by different variables

### LINKED ISSUE #194: Allow ordering of lines by scenario level of increase and health economic value
at the moment we select lines by top no. of cyclists but we could select by top health economic gain (so giving more weight to longer routes and ones replacing fewer pedestrians). This could be an alternative. 

This needs consideration but want to flag it for now

nikolai-b:
Closes #194


usr110:
Great work!

PR #269, opened 2016-06-02
Add centroids layer as a checkbox

usr110:


Robinlovelace:
I think we should test a lower opacity option first as it does not add to the UI. Thanks for creating this for testing - will wait for feedback from others but my 1st impression is that simply reducing opacity is the best way to solve the problem. Great work in in any case.

Robinlovelace:
Closing as superceded by this https://github.com/npct/pct-shiny/pull/274

PR #282, opened 2016-06-16
Update font colour of 'change in death/yr' when its value is negative…

### LINKED ISSUE #271: increase deaths and health economic value
if there is an increase in deaths and thus a negative monetary value then please change font for that line in the label to red. see example
[example.docx](https://github.com/npct/pct-shiny/files/297324/example.docx)

usr110:
…. Fix #271.


Robinlovelace:
Test. Works. Merged.

Fantastic work @usr110 !
selection_157

Issue #276, opened 2016-06-06
Improving the zone/lines data download interface and content


AnnaGoodman1:
Here are the steps I think I needed for this issue:

1) change the “here you can download” text on the screen to match that shown in the attached word document, and also increase the font size.  Note that we are now removing some of the CSV files that could previously be downloaded

2) change the variables included in the CSV file and RDS/Geo files, and the ordering of these variables, to match those specified in the attached word document and in the attached Excel document.   

3) change the variables displayed on the screen, and their order and the labels given to them, to match those specified in the attached Excel document

4) in a few cases variables need to be altered or added or renamed (or if renaming is not possible, then they should be deleted). These cases are highlighted in the first columns of the attached Excel document.  (There are also other inconsistencies in the naming which I find unhelpful – e.g. the car driver variable is "Car" in the zones, and "Car_driver" in the lines dataset; there is inconsistency between the files as to whether variable names are capitalised, e.g. “foot” in the zone file and “Foot” in the line file.  I previously asked Robin if these can be harmonised, but his view was that it would be too much work to be worth it)

5)  the attached Excel document will form the basis of a zones and a lines codebook that I will create once the above changes have been put in place – just in case anything needs to change.

I hope this makes sense, but please ask me if not then I can try to explain further. I’m afraid you would need to ask me any questions by email, since I don’t want monitor github

Anna

[downloadzoneslines.zip](https://github.com/npct/pct-shiny/files/300134/downloadzoneslines.zip)

AnnaGoodman1:
Thanks for the work you have been doing on this, Nikolai - I have just been having a look and taking stock.  Here are the further steps I think I needed for this issue (Some of these are just repeating what I said in my first message, so that everything is in one place – others are in response to the work you have already done):

1) change the “here you can download” text & download options on the screen to match that shown in the attached word document - Specifically, the parts highlighted yellow have not yet been implemented.  Note that we are now removing some of the CSV files that could previously be downloaded.  

2) zones: 1) change the variables included in the CSV file and RDS/Geo files, and (Ideally) the ordering of these variables, to match those specified in the attached Excel document.  

3) Lines: 1) change the variables included in the CSV file and the straight lines RDS/Geo file, and (Ideally) the ordering of these variables, to match those specified in the attached Excel document. 2) The RDS/Geo files of the fast and quiet routes should not show any attribute data (but should have ID in there, ie start and  end  zones). 3) The route network file should continue to contain scenario data, it does not need to be changed.

[...]

PR #309, opened 2016-07-08
Zone and line renaming

nikolai-b:
Ref npct/pct#23 and ref npct/pct-bigdata#1 and ref npct/pct-shiny#276

Using lowercase names. Not ready for merging yet as need to remove csv download options for rnet et. .al.


server-base.R
```diff
@@ -224,7 +224,7 @@ shinyServer(function(input, output, session){
   observe({
     region$replot
     input$map_base
-    show_zone_popup <- isolate({input$line_type == 'none'})
```
usr110: hey @nikolai-b, may I ask why you removed isolate from here?

Robinlovelace: (Nikolai on Robin's PC)
Because it didn't work :cry: it meant was always false.  Have a check, when you select lines (or routes) then the background zone should stop being selectable.

usr110: @nikolai-b 
Have double-checked it. At my end, it's not always `false`. This is true when `none` is selected (Cycling Flows) - and zones are clickable. Otherwise, zones are not clickable.

Robinlovelace:
Thanks for the input @usr110 . Would re-adding it fix the issue?
Cheers.

usr110:
@Robinlovelace Thanks for the quick response. Yes, this particular issue will certainly be fixed.

PR #382, opened 2016-08-22
Test new colours

Robinlovelace:


Robinlovelace:
Merging for testing for now - will revert if there are any issues.

PR #393, opened 2016-08-26
Download on fly

nikolai-b:
This is not ready yet but the idea being for the server to generate the download files.  Could be done in a background thread.

The code joins routes and lines but we agreed to not do so I'll remove that.

### DIFF
--- a/pct-shiny-funs.R
+++ b/pct-shiny-funs.R
@@ -3,12 +3,10 @@ sc_name_df <- data.frame(
   sc_s_name = c("olc","govtarget","gendereq","dutch", "ebike")
 )
 
-signif_sdf <- function(sdf, name, digits = 3) {
+signif_sdf <- function(sdf, digits = 3) {
   nums <- vapply(sdf@data, is.numeric, FUN.VALUE = logical(1))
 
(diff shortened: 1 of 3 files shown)


nikolai-b:
@usr110 and @Robinlovelace thoughts on this one?  It means that we can remove 
> * To get attribute data, use 'ID' field to merge with straight-line CSV file 

and also we can tweak and change the fields in the download as much as we want without any rebuilding.  The downside is server load but that seems minimal (locally large regions do not cause a large CPU spike).

Robinlovelace:
Will give it a test - if we merge to master is it worth doing some tests on memory/cpu use?

nikolai-b:
This has been around for a while.  I'm going to merge it tomorrow unless there are objections.

Robinlovelace:
No objections from me - can you do some analysis of performance of it on the test server when it's merged though? I suggest we pick a time, e.g. midday this Friday, for everyone to download lots of data to see how it responds. Let me know if you're free to work on it then also.

nikolai-b:
Maybe, [Apache Benchmark](http://httpd.apache.org/docs/current/programs/ab.html) should probably be able to do this for us.  Once its on the test server I can hammer it and see how it responds.

Issue #228, opened 2016-05-13
Change regional boundaries


nikolai-b:
There should be a Combined Authority for the North East. This covers Northumberland, County Durham and Tyne and Wear.

Moved from https://github.com/npct/pct-load/issues/7

jf1:
The latest ONS data include Combined Authority information.
Available [here](https://geoportal.statistics.gov.uk/geoportal/catalog/search/resource/details.page?uuid=%7B1AC0D632-38B9-4E08-8AC8-6D87DCF10421%7D) and [here](https://geoportal.statistics.gov.uk/geoportal/catalog/search/resource/details.page?uuid=%7B5F424985-A5B2-49F8-AB2C-20CE9E64691C%7D). They're only small files so pasted below...

CAUTH16CD,CAUTH16NM
E47000001,Greater Manchester
E47000002,Sheffield City Region
E47000003,West Yorkshire
E47000004,Liverpool City Region
E47000005,North East
E47000006,Tees Valley

LAD16CD,LAD16NM,CAUTH16CD,CAUTH16NM
E08000001,Bolton,E47000001,Greater Manchester
E08000002,Bury,E47000001,Greater Manchester
E08000003,Manchester,E47000001,Greater Manchester
E08000004,Oldham,E47000001,Greater Manchester
E08000005,Rochdale,E47000001,Greater Manchester
E08000006,Salford,E47000001,Greater Manchester
E08000007,Stockport,E47000001,Greater Manchester
E08000008,Tameside,E47000001,Greater Manchester
E08000009,Trafford,E47000001,Greater Manchester
E08000010,Wigan,E47000001,Greater Manchester
E08000016,Barnsley,E47000002,Sheffield City Region
E08000017,Doncaster,E47000002,Sheffield City Region
E08000018,Rotherham,E47000002,Sheffield City Region
E08000019,Sheffield,E47000002,Sheffield City Region
E08000032,Bradford,E47000003,West Yorkshire
E08000033,Calderdale,E47000003,West Yorkshire
E08000034,Kirklees,E47000003,West Yorkshire
E08000035,Leeds,E47000003,West Yorkshire
E08000036,Wakefield,E47000003,West Yorkshire
E06000006,Halton,E47000004,Liverpool City Region
E08000011,Knowsley,E47000004,Liverpool City Region
E08000012,Liverpool,E47000004,Liverpool City Region
E08000013,St. Helens,E47000004,Liverpool City Region
[...]

PR #410, opened 2016-09-30
Region gen

nikolai-b:
I think we should have the scripts that generate the files used in this repo.

I'm not sure where the [regions.geojson](https://github.com/npct/pct-shiny/blob/aedf25e2e9/regions.geojson) in the root of this repo came from in the first place but I've removed the pointless variables (url, url_text, Upper and pcycle) so it is bare.
The [regions.geojson](https://github.com/npct/pct-shiny/blob/95a051f4e44501aeda6d9baa79156c957b29bfa7/regions_www/regions.geojson) in the regions_www folder is generated an has the additional cycling variables and I'm adding a script to update it.


usr110:
Great!
Originally it comes from [pct-bigdata](https://github.com/npct/pct-bigdata/blob/master/regions.geojson)

nikolai-b:
I don't think it does.  Surely it was just a download from a site or it was generated and I would like to know from where.

usr110:
Oh yes, now it is clear. You're right. The underlying script must be known to @Robinlovelace, as he created it.

Robinlovelace:
No they are custom regions that I created by merging together different LAs. The scripts will be in PCT but not 100% sure where now. As per #228 these regional boundaries are likely to change. Not sure why you've merged this back in now @nikolai-b as before you wanted it out of pct-shiny. Also it's self referential and just repeats what was in https://github.com/npct/pct/blob/master/analysis/mapgen.R so don't.

nikolai-b:
Lets delete https://github.com/npct/pct/blob/master/analysis/mapgen.R then?

In my mind anything that is working code and creates files that are used in this repo should either be in this repo or pct-load.  https://github.com/npct/pct-data is like a database and has no code only data.  The https://github.com/npct/pct repo is far to confusing for me to understand, if we want that to be a place for working code to live then it needs tidying and a massive cull, moving code there is the same as it being deleted as I'm unable to navigate it on my own.

Robinlovelace:
So, pct is about analysis and one-off scripts for generating data. pct-shiny is the interactive web app. The app does not use that code, it's irrelevant to other users and other countries, so should be deleted from pct-shiny. That was exactly the point you were making here:   I find it strange that you want to re-add it now as I previously re-added it after you deleted it!

Here's where you deleted it: https://github.com/npct/pct-shiny/commit/6e4885617abeae144ca811bbecb6dc9f23c72280

Here's where I re-added it: https://github.com/npct/pct-shiny/commit/17ef1258285484cee6db6bdef4efedb6cec078d4

So let's delete it here and job done - going round in circles is not productive.

nikolai-b:
Code that is required to make this repo work should be in this repo (or pct-load), pct is too confusing for me.  I deleted it code that generated the old front page, I added code that updates the regions.geojson so when we change the scenario numbers then it can be regenerated.

Most things are irrelevant to users in other countries so that is a strange point: MSOAs, LSOAs, our images and funders, should we delete them also?

Robinlovelace:
No because they are used by the app. the app does not process mapgen.R. Happy for this to be moved into pct-load and for regions.geojson to live only in pct-shiny. Please update all references to regions.geojson in pct load if you do this though.

PR #411, opened 2016-09-30
Remove old files

nikolai-b:
Removing old assets and old code.
I don't think we are using any of these images (I searched the codebase).  I also don't think we are using the manual Rmd anymore so I've removed that.


static/manual_body.Rmd
```diff
@@ -1,237 +0,0 @@
----
```
usr110: This script generates the manual_body.html file, so I don't think you can delete it.  (this is where all the recent changes in manual were carried out).

Robinlovelace: True that - good to clean up the repo though.

nikolai-b: Thanks, I've kept that `Rmd` file.  It looks pretty static so I thought it was just html, apologies.

usr110: It's alright. Robin might not like the idea, but I believe we don't know the markdown document any more. HTML will serve the purpose fairly well.

nikolai-b: Do you mean don't *need*?

usr110: Corrected, thanks - edited the message.

Robinlovelace: The reason for using .Rmd, aside from making it easier for others to edit (as stated by @jf1 ), is that we can show how to reproduce the results. Writing this html from scratch would take a long time!

manual-eg

PR #414, opened 2016-10-03
Update production branch

usr110:


Robinlovelace:
Good idea - we've made lots of improvements that are worth pushing.

PR #417, opened 2016-10-04
Generic switch data layer

usr110:
Introduce mechanisms to load `all-trip` data for a region, where it exists. Allows users to switch from `commute` to `all` trip data using a `radio button`.
Hopefully by tomorrow the updated data for Greater-Manchester (for `all-trip`) will be pushed onto the `pct-data` by @audev.


usr110:
If you want to test the functionality out, try running the app for the `greater-manchester` region, as it's the only region with `all-trips` folder.

Robinlovelace:
Great work Ali, will test by Friday.

server-base.R
```diff
@@ -68,11 +71,39 @@ shinyServer(function(input, output, session){
...
+    region$repopulate_region <<- F
+    #isolate(region$replot <- !region$replot)
+  })
+
+  region <- reactiveValues(current = starting_city, data_dir = file.path(data_dir_root, starting_city), repopulate_region = F) # replot = F,
 
-    isolate(region$replot <- !region$replot)
+  observe({
+    # If a region does not have an 'all-trips'directory, disable the dropdown menu
+    if (!dir.exists(file.path(data_dir_root, starting_city, 'all-trips'))){
```
nikolai-b: Your doing this check in a few places.  Maybe you could store it as a reactive value on `region` here,
e.g.
```R
region$all_trips <- dir.exists(file.path(data_dir_root, starting_city, 'all-trips')
```
then check for that in other places?

server-base.R
```diff
@@ -172,7 +203,16 @@ shinyServer(function(input, output, session){
...
-    new_data_dir <- file.path(data_dir_root, new_region)
+    # Check if the new_region is not null, and contains 'all-trips' subfolder
+    if (!is.null(new_region) && file.exists(file.path(data_dir_root, new_region, 'all-trips'))){
+      if (input$trip_type == 'All'){
+        new_data_dir <- file.path(data_dir_root, new_region, 'all-trips')
+      }else{
+        new_data_dir <- file.path(data_dir_root, new_region)
+      }
+    }else
+      new_data_dir <- file.path(data_dir_root, new_region)
```
nikolai-b:
```R
new_data_dir <- if (region$all_trip && input$trip_type == 'All'){
                  file.path(data_dir_root, new_region, 'all-trips')
                } else {
                  file.path(data_dir_root, new_region)
                }
```
Not sure if that's valid R but do you see what I'm suggesting?

server-base.R
```diff
@@ -68,11 +71,39 @@ shinyServer(function(input, output, session){
...
+    # If it does, only then load 'all-trip' data or load defaul commute data
+    if (dir.exists(file.path(data_dir_root, starting_city, 'all-trips'))){
+      if (input$trip_type == 'All'){
+        region$data_dir <<- file.path(data_dir_root, starting_city, 'all-trips')
+      }
+      else{
+        region$data_dir <<- file.path(data_dir_root, starting_city)
+      }
+      # redraw_zones()
+      region$repopulate_region <<- T
```
nikolai-b: Does this need to be set to false when you change region?  Not an issue now as GM is isolated.

server-base.R
```diff
@@ -544,16 +584,9 @@ shinyServer(function(input, output, session){
...
-  shinyjs::onclick("toggle_legend", shinyjs::toggle(id = "zone_legend", anim = FALSE))
+  shinyjs::onclick("toggle_trip_menu", shinyjs::toggle(id = "trip_menu", anim = FALSE))
   shinyjs::onclick("toggle_map_legend", shinyjs::toggle(id = "map_legend", anim = FALSE))
 
-  observe({
-    if (input$map_base == 'IMD')
-      shinyjs::hide("zone_legend")
-    else
-      shinyjs::show("zone_legend")
-  })
```
nikolai-b: Looks like there is a window clash on IMD
screenshot from 2016-10-04 21-34-38

usr110:
Thanks for the feedback, Nikolai. I will have a close look tomorrow.

server-base.R
usr110: Yes, I do. Thanks. Will make the update.

server-base.R
usr110: When a user moves to a new region, it is set to `false`. See here:
https://github.com/npct/pct-shiny/blob/generic_switch_data_layer/server-base.R#L74

Re isolate, it won't be there once the integration happens - will remove the file.

server-base.R
usr110: Good point. Will shift the radio button to the right side.

usr110:
Re https://github.com/npct/pct-shiny/pull/417#pullrequestreview-2795159

After the integration, we won't have either `gm-all-trips` or `greater-manchester-NC` folder; both of them will be deleted.

PR #426, opened 2016-10-21
Update production with layer switching abilities

Robinlovelace:
@nikolai-b @usr110 let's update the production server.

Any reasons why not?

Robin


usr110:
Sure. Especially with the missing map base, it's a good idea.

Robinlovelace:
Thanks Ali.

PR #430, opened 2016-10-31
All trips tidy

nikolai-b:
@usr110 what do you think?


Robinlovelace:
The generic title is a definite improvement. What's the thinking behind `repopulate_region`?

Issue #436, opened 2016-11-12
Data table search/order functionality is not working


Robinlovelace:
Reproducible example:

Click on on the lines tab in the PCT for a region, e.g. http://www.pct.bike/west-yorkshire/ and select the desire lines layer.

Click on "Total no. commuters". This happens - fortunately this does not break the app, leading to a restart but is definitely a bug. Any ideas @nikolai-b or @usr110?

![screenshot from 2016-11-12 12-12-39](https://cloud.githubusercontent.com/assets/1825120/20237929/9a18975c-a8d7-11e6-9e00-196610ea3111.png)

Fixed on test server so will update production - thanks for fixing @nikolai-b if it was you!

Issue #435, opened 2016-11-11
Test server crashing (on GM layer)


audev:
I've just pushed the latest build for the GM layer to the test server.
http://npct0.vs.mythic-beasts.com/greater-manchester/

Last week all worked fine (aside from some connection issues esp. on Firefox). 
Now it seems to crash when:

- The Trip data mode is set to "All"
- An action is triggered

@nikolai-b  Could this be linked to the changes in labels for the layer? I don't see any other relevant changes recently.

JDWoodcock:
please can @nikolai-b or @Robinlovelace have a look at this as we would like to show new layer in Manchester on Monday

nikolai-b:
I struggling to re-create this issue.  I believe it my well be a dup of https://github.com/npct/pct-shiny/issues/433#issuecomment-259302479 which has been fixed a day ago in https://github.com/npct/pct-shiny/commit/31f67f2b4bf25fbff41bd9eb76c85649ede69653.  Please can you confirm if this is still an issue and the steps required to re-create it?

On a side-note http://npct0.vs.mythic-beasts.com is not designed to be reliable, it is designed to be a test server.  Strongly suggest we should only be showing / demoing things on www.pct.bike as that is more stable and should never be broken.  If we are happy with the state of the test server (npct0) then lets deploy it to www.pct.bike today!

JDWoodcock:
thanks for looking when I go to all trips and select lines it crahses and disconnects from server
[test.docx](https://github.com/npct/pct-shiny/files/587340/test.docx)
As I can't get it to work I dont know if I am happy with it or not! However, even if the numbers look good there would be much more work before we could put it on the live server. this demoing is to a tech group of experts- to get feedback on what we have done- it is not close to being shown to public

nikolai-b:
Thanks, that makes sense why it is not on production.
The issue is due to the lines data missing a CRS, I've put in a hacky fix.

There is a separate problem that if you visualize the commuting lines then change to All trips the commuting lines are still shown.  The routes and route network work.  Let me know if this is works as expected.

JDWoodcock:
thanks but I am still getting it crashing when I try and see lines on the 'all-trips' layer

audev:
Also crashing for me. If it's something about the build, I can rebuild any time; but not sure what the CRS you mention is.

audev:
Displaying fine now, thanks @nikolai-b

nikolai-b:
[...]

PR #437, opened 2016-11-12
Update production

Robinlovelace:


Robinlovelace:
Please @nikolai-b or @usr110 accept this PR if it seems good to go.

It has some bug fixes - e.g. https://github.com/npct/pct-shiny/issues/436 - so keen to update pct.bike.

Robinlovelace:
Bump @nikolai-b

nikolai-b:
Do we want all trips on production?
https://github.com/npct/pct-shiny/issues/435#issuecomment-260134143 @JDWoodcock thinks there is much more work before it should go live.  Also doesn't @AnnaGoodman1 have quite few questions #439 about the new data?

JDWoodcock:
no please do not push all trips to production ! @Robinlovelace @nikolai-b @audev

JDWoodcock:
no please do not push all trips to production ! @Robinlovelace @nikolai-b @audev

nikolai-b:
Actually it looks like this has happened 25 days ago https://github.com/npct/pct-shiny/pull/426

http://www.pct.bike/greater-manchester/ has all trips.

@JDWoodcock if we really don't want in on production we can hide it on testing, merge that into production and then re-enable it.  It would be very simple.  Are we happy with the new builds going onto production?

JDWoodcock:
@nikolai-b @nikolai-b please remove it from the production version as a  matter of urgency.
@AnnaGoodman1 @RachelAldred 
I have not checked the new builds, but they definitely need to be checked before they go live

Robinlovelace:
@JDWoodcock @audev you need to remove the GM layers files from the production branch. I suggest a separate `gm_layers` branch is set-up to avoid this happening - we have been working in this way, testing updates on `master` and pushing to `production` incrementally for a long time so I assumed when these updates were pushed the desire was for them to go into production as has happened with all other updates pushed to `master`. Happy to look at this but it won't be until Wednesday so suggest @audev or @nikolai-b does it if either get a spare moment.

JDWoodcock:
@Robinlovelace I understand that we agreed that someone from the academic team in addition to you has to sign things off before they are pushed

PR #503, opened 2016-12-21
Update production with recent changes

Robinlovelace:
Heads up @usr110 @JDWoodcock @RachelAldred @AnnaGoodman1.

This will implement new URL switching scheme.

Robin


nikolai-b:
@Robinlovelace please don't merge this till I've updated the servers to use the new config.

nikolai-b:
- Change the URLs: 

| Old | New | 
|-|-|
| http://npct0.vs.mythic-beasts.com/avon | http://npct0.vs.mythic-beasts.com/msoa/?r=avon  | 
| http://pct.bike/cambridgeshire | http://pct.bike/msoa/?r=cambridgeshire | 

There is a redirect so visiting an old style URL will automatically forward you to the new style URL
- Fix missing line and zone data when switching regions
- Fix strangeness around changing all trips (all trips is still not available on pct.bike)

mvl22:
> http://pct.bike/cambridgeshire

Shame to lose such a nice URL!

AnnaGoodman1:
Thanks for this Nikolai.  I am happy with the switch to ?r=cambridgeshire, I am less sure about the addition of /msoa/ to the URL since we are about to be launching that contains (and can jointly show) a mixture of MSOA and LSOA data - I hadn't appreciated at the last meeting that that addition was also proposed.  (if anything the more relevant condition might be something like '/commuting/', but I'm also not sure that is a good idea, since ultimately we may want to show  meeting in schools by the same URL using switching). I appreciate that it's good to get this out at quiet time, but probably we ideally need to hear from James and Rachel before it pushes to production ...

nikolai-b:
Can we keep it shorter than commuting please? Zones are still MSOA but when you are talking about LSOA data do you mean a the LSOA base maps?

For the new URLs how about:
`/map/?r=avon`
or
`/m/?r=avon`

> we ideally need to hear from James and Rachel before it pushes to production ...

Fine, we can wait till @JDWoodcock and @RachelAldred are back, no big rush.

Robinlovelace:
I think region urls like http://pct.bike/cambridgeshire should still work - is it the plan @nikolai-b to redirect that to http://pct.bike/msoa/?r=cambridgeshire ? I thought that was the plan and think it would be a sensible default.

nikolai-b:
@Robinlovelace https://github.com/npct/pct-shiny/pull/503#issuecomment-268974508
> There is a redirect so visiting an old style URL will automatically forward you to the new style URL

Robinlovelace:
Cheers Nikolai, just double checking. No worries @mvl22 you won't lose your nice http://pct.bike/cambridgeshire url!

AnnaGoodman1:
@nikolai-b , I agree 'commuting' is too long as well as likely to become outdated.  Is 
'http://pct.bike/?r=avon' not an option?'/map/?r=avon' does not seem ideal either unless there will be a different url for the download pages.... .  

Or if we just need something there, doesn't matter what, then maybe 'http://pct.bike/england/?r=avon' or 'http://pct.bike/uk/?r=avon'

nikolai-b:
`m` made sense to me because the downloads are linked to the map (and the map is the primary interface).  Yes it does need something so sadly http://pct.bike/?r=avon is not an option.
I'm not a fan ok `uk` because we don't have the UK in there yet (e.g. Scotland, Northern Ireland).  I'd prefer `m` or any other letter of the alphabet, it doesn't have to mean anything, look at amazon's URL for kindle products e.g.

https://www.amazon.com/b/ref=br_pdt_mgUpt/151-1225585-6231519?_encoding=UTF8&node=6669702011&pf_rd_m=ATVPDKIKX0DER&pf_rd_s=&pf_rd_r=BJQ4YMJW8AKF7XFW05RN&pf_rd_t=36701&pf_rd_p=9974eadb-3306-415d-bd0b-c251318255d0&pf_rd_i=desktop

it starts with a `/b/`

Robinlovelace:
FWIW I like MSOA as the results are primarily at the MSOA level.

mvl22:
Is there a proper issue where the URL stuff can be discussed?

URLs are something I always give a *lot* of thought to when designing systems, and I'd be interested to give some thoughts.

Robinlovelace:
Views welcome but we cannot guarantee to heed them. Here's as good as anywhere as this is the PR in which the changes will go into production so appropriately 'close to the metal'

A table with each option and pros and cons could be useful, e.g.

|Option |Pros |Cons |
|:--|:--|:--|
|msoa |NA |NA |
|m |NA |NA |
|something-else |NA |NA |

AnnaGoodman1:
if there has to be something there then I would prefer 'm' to 'msoa'.

If we did have something that meant something - which if poss might be nice, I suggest 'eng' or 'e' for 'england', or 'ew' for 'England and Wales' (I take the point that we don't have all of the UK there yet, but we might hopefully have Wales before too long, so if we didn't go for 'uk'or 'ew' we would be having to plan that the Wales extension to England would have a different little bit there, e.g. '/wal/').  Probably 'ew' would work out easier in the long run.

So my preferences are 'ew'  > 'eng' > 'm' > 'msoa'.  But I could live with any of them.

Robinlovelace:
Thanks Anna - I realise that I initially said msoa > m but at the minute it's m ~ msoa as they both sound reasonable. I think ew would also be a strong contender. @JDWoodcock and @RachelAldred please provide input as it seems we don't have a strong 'winner' yet.

RachelAldred:
I think my preferences are eng > ew > m > msoa, on the grounds that we don't have Wales yet (but I see the benefit of having it longer term). But also could live with any of them.

RachelAldred:
Ah, I didn't realise m was map rather than short for msoa! In which case eng > m > ew >msoa, but happy with all still.

usr110:
Nikolai, was the change in URL introduced to deal with the data table issues? Sorry but I am still not clear why it was changed. If you can expand on it, that would be great.
If it's possible I too would prefer `pct.bike/greater-manchester` or `pct.bike/?r=greater-manchester`
For a test, I moved the content of `msoa` to one level down, and locally it seems to work fine. This is how the (local) URL looks like:
http://127.0.0.1:6911/?r=essex

Robinlovelace:
To clarify: `pct.bike/greater-manchester` would still work but would re-direct. I think the main benefit of not having it as the base url will allow us to distinguish between different versions/places - `eng` would emphasise the latter which would be useful. What are the advantage of using simply `/?r=region` as opposed to `/m/?r=region` or `/eng/?r=region` or `/somethingelse/?r=region` @usr110?

usr110:
Redirection means that it won't be visible to users, so I don't think that's useful. It will only work as compatibility for those who've saved the old URLs.
Re the other point, as said by @mvl22 and @AnnaGoodman1 it is a good practice to keep the URL simple. Now that @nikolai-b has explained to me on slack why we need an additional folder (such as `eng`), I am happy with any solution - whether it's `m` or `eng` or whatever is decided.

mvl22:
> Redirection means that it won't be visible to users, so I don't think that's useful.

+1. A non-visible URL is basically non-existent.

Robinlovelace:
Clarification: redirection is just to ensure compatibility with existing URLs such as http://pct.bike/cambridgeshire 

Update: at the insistence of @mvl22, I've created an issue for this which can be closed when we decide on a solution: https://github.com/npct/pct-shiny/issues/504

Great - sounds like there are no outstanding objections to introducing a sub-folder. Now we just need to decide which one. Please cast your vote here: http://doodle.com/poll/a2ggpwy3d8r8h93i#table

AnnaGoodman1:
@nikolai-b , I have spoken with Robin, Rachel and James and the consensus is that we would prefer 'eng' instead of 'msoa' to be the placeholder in the middle of the PCT URL (e.g. http://npct0.vs.mythic-beasts.com/eng/?r=staffordshire).  Can I check, if we do this and subsequently add Wales would it be possible to have a different placeholder (/wales/) for the regions in Wales?  If so then would you be able to change the /msoa/ to /eng/ at some point please? I think at that point we would also be happy for this to go live on the main server.

Sorry that it has been a slightly laborious process to get here...

nikolai-b:
@AnnaGoodman1 we would be able to use `/eng/` and `/wales/` but if we do then we will not be able to switch between `/eng/` and `/wales/` regions, just wanted to make that clear before we push ahead.

Robinlovelace:
@nikolai-b @mem48 @usr110 @RachelAldred @JDWoodcock @AnnaGoodman1 I think we should do this asap. What still needs to be done? On my list:

- [ ] raster tiles (can be done after, does not affect pct-shiny)

That's it. So I think we can push to production now.

Comments here if you think otherwise or would like one more change to get in there.

JDWoodcock:
thanks but can we see a summary of the changes you are planning to push before they are pushed

Robinlovelace:
All of the changes discussed and mentioned here and shown on the test server. Summary:

- Bug fixes
- LSOA route network layer
- Updated url

JDWoodcock:
As agreed at Friday's meeting we will not push the LSOA layer until it has had some user testing and we have the manual.

Robinlovelace:
True that. Hence double checking.

PR #533, opened 2017-02-21
Release

nikolai-b:
Change log:

- Move to the new URL structure
- Fix line and zone tables when a moving between regions (and a few other region changing issues)
- Fix some issues with all_trips
- Add new user manuals and case studies
- Add Malcolm to the team page
- Hiding zones hides the boudaries and stops popups
- Add references to the paper


usr110:
If this `pull-request` is to update the `production` branch then @nikolai-b, just to make sure, `all-trips` should not be there. Any kind of regional prototypes should be excluded. Please confirm @JDWoodcock and @AnnaGoodman1

usr110:
Nikolai, just remembered that I had written code NOT to have `all-trips` on the production branch here: https://github.com/npct/pct-shiny/blob/master/regions_www/m/server.R#L123-L133

We should be OK then

nikolai-b:
@usr110 maybe yes but they are different.  From my understanding we want the LSOA layer to go live very soon (i.e. next release) whereas the `all-trips` I've no idea when that is going live so that needs a long term solution.  If this was more than a slight delay in releasing the LSOA layer then I would agree but not sure it is worth it in this case.

nikolai-b:
@Robinlovelace I've made the production changes so this is ready to go if it has approval.

usr110:
Fair enough @nikolai-b 
Re `all-trips`, as I mentioned earlier, no region-specific change would go live.

JDWoodcock:
this also applies to near market scenario too.

Robinlovelace:
Many thanks @nikolai-b for setting this up and making the necessary adjustments on the production server. Last opportunity @JDWoodcock @RachelAldred and @AnnaGoodman1: what you see on the test server, but without the LSOA rnet is what I plan to push.

**By midday tomorrow.**

nikolai-b:
@Robinlovelace sorry I set the config up on npt1 but not on npt2.  Also there was a strange error from CircleCI https://circleci.com/gh/npct/pct-shiny/691
not sure why.  Should be all live and working now.

PR #539, opened 2017-03-05
535 allow user to confirm region change

nikolai-b:
Very basic demonstration of region change based on mouse click.

Issues are 
1. The popup obscures the mouse click (e.g. sometimes you click on the popup which doesn't change the region instead of the background which does)
2. The region name is lower case and not pretty
3. (Nice to have) feedback that the region change is happening, esp for large regions which take a second or two to load.

All these are fixable but I'd like general feedback first.

### DIFF
--- a/regions_www/m/server.R
+++ b/regions_www/m/server.R
@@ -114,6 +114,7 @@ shinyServer(function(input, output, session){
     region$all_trips <- dir.exists(file.path(data_dir_root, region$current , 'all-trips'))
 
     region$repopulate_region <- T
+    leafletProxy("map") %>% removePopup(., "new-region") %>% removeShape(., "new-region-outline")
   })
 
   region <- reactiveValues(current = NA, data_dir = NA, repopulate_region = F, all_trips = NA)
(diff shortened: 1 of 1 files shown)


Robinlovelace:
Just tested this - the region name you're over pops-up:

image

I like the fact you can go straight to anywhere in the country with one click:

image

However this has still got some-way to go under this implementation - these are the outstanding issues I think need resolving before this goes in:

- [ ] The clicking does not always seem to work - to reproduce, try zooming a long way out and clicking on many regions, one by one. Does it switch every time for you? Does not for me.

- [ ] The names need to be capitalised, so Hereford and Worcester - a minor point, we've already solved this in a reproducible way so quite low priority.

- [ ] The pop up does not go away but hangs around in sometime seemingly random places on the screen even when zoomed. I think it would be better if it disappeared here for example:

image

Are there any other implementations that would be easier? How about a popup that you have to click twice on (one to make the popup appear, again to actually switch region)? That would be more explicit and less confusing I think. Good progress towards user-controlled region switching in any case.

nikolai-b:
@Robinlovelace did you see my issues not yet addressed at the top?

> The clicking does not always seem to work - to reproduce, try zooming a long way out and clicking on many regions, one by one. Does it switch every time for you? Does not for me.

is the same as
> 1. The popup obscures the mouse click (e.g. sometimes you click on the popup which doesn't change the region instead of the background which does)

and
> The names need to be capitalised, so Hereford and Worcester - a minor point, we've already solved this in a reproducible way so quite low priority.

is the same as
> 2. The region name is lower case and not pretty

The other point:
> The pop up does not go away but hangs around in sometime seemingly random places on the screen even when zoomed. I think it would be better if it disappeared here for example:

Yes, would be great to make it disappeared but I'm not sure the criteria for that.  Should it be on map zoom, after a period of time...? Might be best if it was a label not a popup, this is still in concept phase so suggestions welcome.  I've made it so it doesn't track the mouse anymore and appears only once per region.

We could make it so you have to click on the the popup to change the region, that is not easy as popups are classed as controls https://rstudio.github.io/leaflet/shiny.html#inputsevents so do not generate mouse events.  I've updated it so the region boundary is visible (we can make this prettier but at the moment it is the default blue) which along with the popup not tracking the mouse seems like an improvement.

Robinlovelace:
Apologies for the duplicate issues - should have read your comments on the PR properly.

I'm very happy with this implementation and am happy for others to test it so going to merge now.

If you could fix the naming thing by tomorrow lunch-time that would allow others to test a solid demo version and get it on production by midweek, closing the issue.

Here's a gratuitous screenshot of jumping from Hereford to Leeds:

image

Robinlovelace:
Also, let me know if you'd like me to fix the naming thing - an easy one I think.

nikolai-b:
Great, thanks for the quick feedback.  I'll get the the name changes done asap.

usr110:
Great work @nikolai-b. Have two points in my head:

1. Would be nice if a user clicks on a new region and then the centre of the map adjusts accordingly. At the moment the map doesn't pan when a user selects a region.
2. More of an aesthetic issue, but with the development branch of `leaflet` (on `GitHub`) one can now have `mouseover` labels. Just redefine the text inside the `label` parameter for the `addPolygons` method. An example is mentioned here: http://rpubs.com/bhaskarvk/polygon-highlight
In the long run, we may update the currently used popups in the app.

Robinlovelace:
Responses from me:

1. I don't think the map should pan: I like that you can be zoomed out and click on many regions without the map jumping, that's useful and less surprising than Ali's (much appreciated) suggestion.

2. Yes - I've flagged these development options, and those in `leaflet-extras` and they make sense to use, many thanks for the link.

I plan to accept this PR and message others for testing imminently so if there are any other issues (neither of those are blockers) let me know asap!

usr110:
Here's my response to point 1:

1. Your point makes sense at a national level. However when the app loads at zoom `10` (and above), intuitively a user expects to move the map to a new region when it's loaded. There's no element of surprise there.

Robinlovelace:
I think there could be surprise either way when zoomed in, especially if it transports you to the centre of an unknown region after you clicked on it out of curiosity without knowing really what to expect. That would be disconcerting for me in any case, and certainly jumpy.

Robinlovelace:
Looking great, merging:

image

usr110:
>I think there could be surprise either way when zoomed in, especially if it transports you to the centre of an unknown region after you clicked on it out of curiosity without knowing really what to expect. That would be disconcerting for me in any case, and certainly jumpy.

Perhaps the wider team might have different views on it. Let's see what they say about it.

AnnaGoodman1:
Thanks all, looking good – I think this will be an excellent improvement to functionality.

I personally think having the map move to centre on the clicked region (keeping the current level of zoom) would probably be a helpful feature, although hard to be sure without seeing it – and not essential if implementing this is hard.

I cannot get London or West Midlands to load by clicking on it, I appreciate that large regions take a few seconds to load but even the URL does not seem to change when I click on London.  This is true in both Firefox and IE

These regional boundaries have been simplified, and don't match the actual regional boundaries we are using, which looks a bit messy. Would it be possible to create regional boundaries directly from the MSOAs, and use those to define the regions so that they align better?  Or at a minimum, could the boundary around each region become thinner/be removed, so the nonalignment is not so obvious? (e.g. be a blue shape but not darker blue border).  Just a thought

Robinlovelace:
Yes that would be possible re more accurate boundaries, good idea that will improve the visual appeal of the outputs. Suggest not allowing the user to jump to London or West Midlands is a higher priority - @nikolai-b is that easy to fix? Maybe worth opening an issue on each of those things Anna. Many thanks for the input.

usr110:
@AnnaGoodman1 I've fixed this issue:

>I cannot get London or West Midlands to load by clicking on it, I appreciate that large regions take a few seconds to load but even the URL does not seem to change when I click on London. This is true in both Firefox and IE

Please test it.

RachelAldred:
It is nice :) Thank you. Small point while using it for London - I find it a little odd that when I am trying to look at London, a neighbouring region remains highlighted even when I am not hovering above it -

image

Robinlovelace:
Good point Rachel - a little disconcerting. Is that an easy fix @nikolai-b or @usr110?

Cannot think of a quick and easy way to solve it but if there is one I'm in favour of your suggestion: https://github.com/npct/pct-shiny/issues/545

PR #547, opened 2017-03-09
Update production: add lsoa layers, user-controlled region switching, lines don't disappear

Robinlovelace:


Robinlovelace:
@JDWoodcock @RachelAldred @AnnaGoodman1 this will fix a raft of issues and finally allow people to make good use of the excellent LSOA layer created by @mem48. 

The only blocker I can see from a tech perspective is that the regions don't re-appear, an issue introduced when @nikolai-b fixed the issue that the region shape persists, as flagged by @RachelAldred.

@nikolai-b are there any other technical issues?

@npct/admin is there anything else preventing this happening?

All the best,

Robin

RachelAldred:
It looks lovely! I guess we have to be a little careful in communicating to people how it works, because it's easy to zoom around the country on the LSOA layer and then end up not being able to see the other layers where you are (as you are still in the other region for the MSOA layers). We may get some people saying e.g. they think it's broken based on that. Not sure if there is a solution to this, though, without losing the zoomability of the LSOA layer.

We do need to update the documentation, and I have drafted some changes to the manual sections. I will now create a new issue with those, for comment. I would really appreciate a few tips ( @Robinlovelace or anyone?)  on using the TIF LSOA network files, as I have so far not been able to open them successfully in QGIS. I'm willing to give advice to users (and I know a number of our stakeholders are using QGIS) but will have to be able to use it myself first!

Finally I think we need to tweak the Model Output page to make it clear that the exclusion criteria apply to the MSOA-level flow data and that the LSOA-level flow data does not for instance exclude cross-regional flows.

RachelAldred:
By the way, sorry, I realise you want to upload the LSOA layers, @Robinlovelace , and I don't think that should happen till we agree the documentation changes (which hopefully will be quick - I'll upload them now for comment).

Robinlovelace:
Many thanks Rachel - now I see the dependency between enabling the LSOA layer and the docs you've been working on. I'll get back to you on that asap - it's in an email I believe. Unfortunately that may not be until early next week unless I find time this evening.

nikolai-b:
There are no outstanding technical issues from me.

RachelAldred:
Hi @Robinlovelace and all, I opened it yesterday a Github issue (the documentation) - it's basically a few paragraphs to add into Manuals B and C, alternatively you can also see them in context in my Dropbox (work in progress/manual)

https://github.com/npct/pct-shiny/issues/548

Robinlovelace:
Heads up @nikolai-b, @usr110, @RachelAldred, @AnnaGoodman1 and @JDWoodcock I plan to accept this around 16:00 today. To me it's looking good-to-go with the new LSOA manuals. Pending rebuilding the model output for all regions (not essential for this) we've fixed all major issues I think. Great work. Input here before we do that welcome.

JDWoodcock:
well done all :)

PR #552, opened 2017-03-14
Update date of JTLU paper

Robinlovelace:


mvl22:
You should link to the doi version, which means that it will continue to work if the journal website gets redesigned:

http://dx.doi.org/10.5198/jtlu.2016.862

Issue #545, opened 2017-03-07
Boundaries and popups do not disappear when not hovering over an outside region


Robinlovelace:
See https://github.com/npct/pct-shiny/pull/539#issuecomment-284777050 for a lucid description of this issue.

nikolai-b:
My fix isn't ideal as when you mouse out of a region then mouse back in it doesn't always re-appear.  I'm guessing its a leaflet issue but :man_shrugging:

Robinlovelace:
This isn't fully fixed so re-opening.

It's intermittent now - not a high priority but worth being aware of in case anyone has any flashes of insight into how to fix it. My own preference would be to have consistency - the blue boundary never disappearing - rather than intermittentent behaviour.

Perhaps it is a leaflet issue, worth putting in an issue at https://github.com/Leaflet/Leaflet ?

Robinlovelace:
My current preference: go back to having non-disappearing boundaries - just set a transparent fill should be a 5 minute job I'd imagine...

Robinlovelace:
Fixed by Ali - now behaves consistently. Very smooth and nice colours, so closing.

![image](https://cloud.githubusercontent.com/assets/1825120/23946234/87f82f06-0971-11e7-8ee6-58a3b51a7e3c.png)

PR #554, opened 2017-03-14
Highlight polygons using leaflet library

usr110:
Instead of adding highlighted region (polygons) on the fly, add all the regions in the start. 

Covers:
- Simplifies the code
- Highlighting is quicker
- Does not show any highlighted region when the mouse is over the already selected region

Needs to do:
- Set the view of the map when the app starts
- Further testing

### DIFF
--- a/pct-shiny-funs.R
+++ b/pct-shiny-funs.R
@@ -3,6 +3,13 @@ sc_name_df <- data.frame(
   sc_s_name = c("olc", "govtarget","gendereq","dutch", "ebike")
 )
 
+get_pretty_region_name <- function(region_name){
+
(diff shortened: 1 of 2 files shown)


usr110:
Relevant to https://github.com/npct/pct-shiny/issues/545

Robinlovelace:
Sounds great Ali, will test asap.

Robinlovelace:
@usr110 is this ready to test?

usr110:
Yes it is. The only issue I am dealing right now is that zooming doesn't need to change when map layer is changed.

usr110:
@Robinlovelace now everything should be in order. Thanks for testing.

Robinlovelace:
This is excellent work Ali. Accepting it into testing.

PR #565, opened 2017-03-24
Update production with fixed base-maps

Robinlovelace:
Planning to accept this PR around 15:00 today unless I hear back from @nikolai-b and @usr110 that there are any tech issues.

Heads up @AnnaGoodman1 @RachelAldred @JDWoodcock this is a 'bug fix' update.


Robinlovelace:
Update: this seems to have worked (see highest zoom level with IMD below): 

image

PR #586, opened 2017-04-11
Update production

Robinlovelace:
Will tag others when this goes ahead.

### DIFF
--- a/regions_www/m/server.R
+++ b/regions_www/m/server.R
@@ -217,6 +217,8 @@ shinyServer(function(input, output, session) {
   to_plot <- NULL
   helper <- NULL
   helper$e_lat_lng <- ""
+  helper$old_geog <- ""
+  helper$old_purpose <- ""
 
   ## Set  values of region
   observe({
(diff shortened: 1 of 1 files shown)


regions_www/la-map.html
```diff
@@ -590,7 +590,7 @@
     var credctrl = L.controlCredits({
       image: "www/android-icon-36x36.png",
       link: "//www.pct.bike",
-      text: "Part of the Propensity <br/>of Cycling (PCT) project"
+      text: "Part of the Propensity of<br/>Cycling Tool (PCT) project"
```
Robinlovelace: That should say the Propensity to Cycle Tool (PCT).

usr110:
Don't know which version you looked into but it already says that @Robinlovelace 

https://github.com/npct/pct-shiny/pull/586/files#diff-ab6b6cde9215e78913a29781184526ffR593

You can double check it on the test-server

usr110:
Robin, do you plan to accept this `PR` ?
I don't see any reason why we can't merge the two branches. This will do the following:
1) Update la-map
2) Update manual page

Robinlovelace:
The regional flows are what's preventing it at the moment. Plan to update it when that's done next week. Cheers for the reminder.

usr110:
Fair enough, Robin. If you are fine with it I will push `la-map` relevant commits onto `production` then.

usr110:
and also `manual' related - see https://github.com/npct/pct-shiny/pull/595#issuecomment-301079299

Robinlovelace:
Happy with that - note this may help:

```bash
man git-cherrypick

NAME
       git-cherry-pick - Apply the changes introduced by some existing commits

SYNOPSIS
       git cherry-pick [--edit] [-n] [-m parent-number] [-s] [-x] [--ff]
                         [-S[<keyid>]] <commit>...
       git cherry-pick --continue
       git cherry-pick --quit
       git cherry-pick --abort

DESCRIPTION
       Given one or more existing commits, apply the change each one introduces, recording a new commit for each. This requires your working tree to be clean (no
       modifications from the HEAD commit).

       When it is not obvious how to apply a change, the following happens:

        1. The current branch and HEAD pointer stay at the last commit successfully made.

        2. The CHERRY_PICK_HEAD ref is set to point at the commit that introduced the change that is difficult to apply.

        3. Paths in which the change applied cleanly are updated both in the index file and in your working tree.

        4. For conflicting paths, the index file records up to three versions, as described in the "TRUE MERGE" section of git-merge(1). The working tree files will include
           a description of the conflict bracketed by the usual conflict markers <<<<<<< and >>>>>>>.

        5. No other modifications are made.
```

usr110:
Great. Thanks for the help - already have been using `cherry-pick` command.
One more thing: we might want to remove all the commits relevant to test the language translation, so it won't be a full merge of `production` and `master` branches.

Issue #592, opened 2017-04-20
Linking back to main tool from LA map


OCSFrancis:
It would be good to have the following on http://www.pct.bike/la-map.html

- A link back to the main tool http://www.pct.bike/ (At the top? Could read "Visit the full Propensity to Cycle Tool"

- A link from the right hand key under the different scenarios: "Find out more about these scenarios" Link to http://www.cedar.iph.cam.ac.uk/resources/evidence/eb-14-englands-cycling-potential/#4scenarios ?

usr110:
Thanks @OCSFrancis for raising the issue.

> A link back to the main tool http://www.pct.bike/ (At the top? Could read "Visit the full Propensity to Cycle Tool"

I don't know if you noticed that there's already an icon on bottom right which takes users to pct.bike.
Did you mean to add another link on top of it?

> A link from the right hand key under the different scenarios: "Find out more about these scenarios" Link to http://www.cedar.iph.cam.ac.uk/resources/evidence/eb-14-englands-cycling-potential/#4scenarios ?

What about a tooltip which shows such info? Or perhaps an expandable pane which has more information.

OCSFrancis:
I didn't spot that square at all. And we shouldn't assume that people who come to this map already know about the PCT, so I think it's worth spelling out. It doesn't have to be at the top - it could be along the bottom where the current icon is.

A tooltip/ expandable pane would be even better - good idea.

usr110:
> ..so I think it's worth spelling out

It already says `Propensity to Cycle Tool (PCT)` there. Can you expand on it?

> A tooltip/ expandable pane would be even better - good idea.

Great, thanks. Had that in my mind. Will work on it.

OCSFrancis:
I'm not seeing that at all. I don't know how to add images on GitHub, so I'll email you as screen grab

nikolai-b:
At @usr110 I believe this is fixed, please re-open if I'm mistaken

usr110:
@nikolai-b haven't completed it yet. Needs to deal with tooltips/pane for additional info for both scenarios and variables. If you get the chance, would you have a look at it? I will provide the required text.

usr110:
In addition to the logo with text at the bottom right position, I've added an information pane (right below the variables) @OCSFrancis. Have a look at on our [test-server](http://npct0.vs.mythic-beasts.com/la-map.html). What do you think? @nikolai-b your inputs would be valueable too.

Robinlovelace:
[...]

PR #593, opened 2017-04-20
Make the PCT link more clearer

nikolai-b:
Refs #592
@usr110 What do you think?


usr110:
@nikolai-b Looks great, thanks. Just one thing I'd add: would it be easier to have a same look and feel as the current text? (black fonts and more spacing between the text and the icon)
Like this:
image

usr110:
@nikolai-b Have another thought: how about if we change the javascript function to display the text as always here: https://github.com/npct/pct-shiny/blob/master/regions_www/www/la-map-resources/leaflet-control-credits.js

Any idea how to go about it?

usr110:
Updated the `js` code here: https://github.com/npct/pct-shiny/commit/f95223781253aa68da008faf0b81c2091bd6f801

We can close this `PR`. What do you think @nikolai-b

usr110:
You can see the change on the [test-server](http://npct0.vs.mythic-beasts.com/la-map.html). If you are happy with it, I will push it to the production branch.

nikolai-b:
@usr110 it is better but now the image isn't a link.  I don't like the link being black text but I can change the spacing in this pull if you prefer.

PR #596, opened 2017-08-22
Draft PCT shiny update

AnnaGoodman1:


.gitignore
```diff
@@ -9,5 +9,10 @@
 .Rproj.user
 README.html
 .RData
-pct-data/
+/regions_www/tabs/download_params_sha_current.rds
```
Robinlovelace: This looks like an absolute file path to me. I suggest this: 

```
regions_www/tabs/download_params_sha_current.rds
```

Right? @nikolai-b

README.md
```diff
@@ -11,15 +11,15 @@ To start:
 
 ```sh
 git clone git@github.com:npct/pct-shiny.git
-git clone git@github.com:npct/pct-data.git --depth=1
+git clone git@github.com:npct/pct-outputs-regional.git
```
Robinlovelace: Great you've updated the README also.

Robinlovelace:
I've tested this and it works with a fix in there so merging.

data_sha
```diff
@@ -1 +0,0 @@
-a1da8ed
```
nikolai-b: This will need to be reflected in:

 https://github.com/npct/pct-shiny/blob/3a06d3c6ffb42fd0ac23dfe71a50f0ce6b656b78/scripts/publish-pct-shiny

or the server will not update using the correct data.

nikolai-b:
Moving the robots file will mean that the site will be hidden to search engines, I assume this is not what we want.  It was in static to and only copied to regions on the test server

regions_www/m/server.R
```diff
@@ -19,804 +19,1071 @@
...
-  colours = c("#9C9C9C", "#FFFF73", "#AFFF00", "#00FFFF", "#30B0FF", "#2E5FFF", "#0000FF", "#FF00C5"),
-  labels = c( "1-9", "10-49", "50-99", "100-249", "250-499", "500-999", "1000-1999", "2000+" )
-)
+## Path directories to load data (expect regional data as a sibling of interface_root)
+interface_root <- file.path("..", "..")
+data_regional_root <-  file.path(interface_root, '..', 'pct-outputs-regional')
+data_national_root <-  file.path(interface_root, '..', 'pct-outputs-national')
+
+## Regional sha file [Anna question: next 7 lines added by me, wasn't clear to me otherwise how read the latest master branch? Or was the intention to set manually?]
+gitArgs_reg <- c("--git-dir", file.path(data_regional_root, ".git"), "rev-parse", "--short", "HEAD", ">", file.path(interface_root, "outputs_regional_sha"))
```
nikolai-b: @AnnaGoodman1 Yes, very much it was meant to be set manually as during development the two weren't always in sync (i.e. when variables were renamed or regions were added one by one we might not want to use the head of the data repository)

regions_www/m/server.R
```diff
@@ -19,804 +19,1071 @@
...
+    region$purposes_present
+    input$purpose
 
+    # Identify region
+    query <- parseQueryString(session$clientData$url_search)
     if (is.na(region$current)) {
-      region$current <- if(isTRUE(query[['r']] %in% regions$Region)){
+      region$current <- if (isTRUE(query[['r']] %in% regions$region_name)) {
         query[['r']]
-      } else if (exists("starting_region")) {
```
nikolai-b: This was to allow us to set the starting region without editing the code, i.e. in a local R session

```R
starting_region <- "lincolnshire"
```

Fine to remove but it has been a useful feature.

Robinlovelace: It's still useful and still works for local testing, as you'll see by running it:

```r
shiny::runApp('regions_www/m')
```

@AnnaGoodman1 set the default to the Isle of Wight which is even more useful than Lincolnshire or (I'm sad to say) West Yorkshire.

usr110:
@Robinlovelace please give ample time before you merge. This has caused the `shiny_interface` application to stop working. As I mentioned before the `lfs` for `pct-outputs-regional` wasn't properly setup. Any update on the server stops the `shiny_interface` app.
@nikolai-b Would you please look at the `shiny_interface` data directory's script to see why it's failing?

Robinlovelace:
Apologies for jumping the gun. As discussed with Nikolai and Anna it's probably best to see this as an opportunity rather than reverting the merge (I looked into that option). The plan is to run with this and try to get lfs, and other issues on the test server, fixed. This should allow the test server to be fixed by the end of the week, in preparation for all these improvements to go live sooner rather than later.

regions_www/m/server.R
nikolai-b: @Robinlovelace 
 
```R
starting_region <- "lincolnshire" # Or anywhere!
shiny::runApp('regions_www/m')
```
Used to start the app in Lincolnshire (or anywhere else we picked).

regions_www/m/server.R
```diff
@@ -19,804 +19,1071 @@
...
-    to_plot$cents <<- readRDS(file.path(region$data_dir, "c.Rds"))
+    # Define region geography, forcing a default in cases where the geography is not available
+    if (input$purpose =="commute") {
+      if (input$geography %in% c("msoa", "lsoa")) {
+        region$geography <<- input$geography
+      } else {
+        region$geography <<- "msoa"
+      }
+    } else if (input$purpose =="school") {
+      if (input$geography %in% c("lsoa")) {
```
nikolai-b: Is there any point in this `if` as it is always setting the `geography` to `"lsoa"`?

nikolai-b: Something like:

```R
purpose_geog <- data.frame(
  "commute" = c("msoa", "lsoa"),
  "school" = c("lsoa"),
  "alltrips" = c("msoa"),
)

if (input$geography %in% purpose_geog[input$purpose]  {
  region$geography <<-  input$geography
} else {
  purpose_geog[input$purpose][1]
}

File: regions_www/m/server.R [hunk 7]
```diff
@@ -19,804 +19,1071 @@
...
+    # Identify that region repopulation has happened
+    region$repopulate_region <<- T
 
-    # Add rqincr column to the quiet data
-    to_plot$quieter_route@data$rqincr <<- to_plot$quieter_route@data$length / to_plot$faster_route@data$length
+    # Identify purposes and geographies available in region
+    purposes_list <- c("commute", "school", "alltrips")
+    region$purposes_present <<- (dir.exists(file.path(data_regional_root, purposes_list, "msoa", region$current)) | dir.exists(file.path(data_regional_root, purposes_list, "lsoa", region$current)))
+    geographies_list <- c("msoa", "lsoa")
+    region$geographies_present <<- dir.exists(file.path(data_regional_root, input$purpose, geographies_list, region$current))
```
nikolai-b: So `region$geographies_present` is now a column vector?

nikolai-b: Also I don't think this needs to be stored in the `region` globally as it isn't used elsewhere.

regions_www/m/server.R
```diff
@@ -19,804 +19,1071 @@
...
-    if (!region$all_trips){
-      # hide trip_panel
-      shinyjs::hide("trip_panel")
-    }else{
-      # show trip_panel
-      shinyjs::show("trip_panel")
+    if (file.exists(file.path(region$data_dir, "l.Rds"))) {
+      to_plot$straight_lines <<- readRDS(file.path(region$data_dir, "l.Rds"))
+    } else {
+      to_plot$straight_lines <<- NULL
```
nikolai-b: I think we could make this a bit DRYer

```R
load_data <- function(to_plot, to_plot_subset, filepath|)
  if (file.exists(filepath)) {
    to_plot[to_plot_subset] <<- readRDS(filepath)
  } else {
    to_plot[to_plot_subset] <<- NULL
  }
)
# then to use it:
load_data(to_plot, "straight_lines", file.path(region$data_dir, "l.Rds"))
```

Robinlovelace: Sounds good. Seeing as this PR has already been merged, are you up for making these changes?

.gitignore
nikolai-b: https://git-scm.com/docs/gitignore

In this case both are valid forms, maybe there is a difference more subtle that I'm unaware of.

regions_www/m/pct_shiny_funs.R
```diff
@@ -0,0 +1,758 @@
+#################
+# Name, number and colour functions
+#################
+
+## Region names [NB copy of this also in pct-scripts/00_setup_and_funs - if modify here, modify there too]
+get_pretty_region_name <- function(region_name, the = T){
+  if (the == T) {
+    region_name <- gsub("isle-of-wight", "the-isle-of-wight", region_name, perl=TRUE)
+    region_name <- gsub("north-east", "the-north-east", region_name, perl=TRUE)
+    region_name <- gsub("west-midlands", "the-west-midlands", region_name, perl=TRUE)
```
nikolai-b: With this:

```R
 get_pretty_region_name("isle-of-wight")
```
It might be easier just to map it:

```R
regions_thes <- c(
  "isle-of-wight" = "The Isle of Wight", 
  "north-east", "The North East",
  "west-midlands", "The West Midlands"
)
if(the == T & !is.na(regions_thes[region_name])) {
  return(regions_thes[region_name])
}
```

regions_www/m/pct_shiny_funs.R
```diff
@@ -0,0 +1,758 @@
...
+  if (purpose=="commute") {scenario_name_df$sc_f_name_commute[scenario_name_df$sc_s_name == scenario_name]}
+  else if (purpose=="school") {scenario_name_df$sc_f_name_school[scenario_name_df$sc_s_name == scenario_name]}
+  else if (purpose=="alltrips") {scenario_name_df$sc_f_name_alltrips[scenario_name_df$sc_s_name == scenario_name]}
+}
+
+# Get names in pop-ups (variable by purpose)
+text_zone_header <- function(purpose) {
+  if(purpose=="commute") {"All residents living in zone"}
+  else if(purpose=="school") {"All children living in zone"}
+  else if(purpose=="alltrips") {"All trips starting in zone"}
```
nikolai-b: Nice, a `switch` might be cleaner:

```R
switch(purpose,
  "commute" = "All residents living in zone",
  "school" = "All children living in zone",
  "alltrips" = "All trips starting in zone"
)
```

regions_www/m/pct_shiny_funs.R
```diff
@@ -0,0 +1,758 @@
...
+  sc_s_name = c("olc", "govtarget","gendereq","dutch", "ebike"),
+  sc_f_name_commute = c("Census 2011 Cycling", "Government Target", "Gender equality", "Go Dutch", "Ebikes"),
+  sc_f_name_school = c("School Census 2011", "Government Target", "Gender equality", "Go Dutch", "Ebikes"),
+  sc_f_name_alltrips = c("Current travel patterns", "Government Target", "Gender equality", "Go Dutch", "Ebikes")
+)
+
+get_scenario_name <- function(scenario_name, purpose){
+  if (purpose=="commute") {scenario_name_df$sc_f_name_commute[scenario_name_df$sc_s_name == scenario_name]}
+  else if (purpose=="school") {scenario_name_df$sc_f_name_school[scenario_name_df$sc_s_name == scenario_name]}
+  else if (purpose=="alltrips") {scenario_name_df$sc_f_name_alltrips[scenario_name_df$sc_s_name == scenario_name]}
```
nikolai-b: Again, a `switch` might be better.

regions_www/m/pct_shiny_funs.R
```diff
@@ -0,0 +1,758 @@
...
+  else if(purpose=="school") {"Total children: &nbsp;"}
+  else if(purpose=="alltrips") {"Total weekly no. trips: &nbsp;"}
+}
+
+text_cycle_baseline <- function(purpose){
+  if(purpose=="commute" | purpose=="school") {"Cyclists (baseline): &nbsp; "}
+  else if(purpose=="alltrips") {"Cycle trips/wk (baseline): &nbsp;"}
+}
+
+text_cycle_scenario <- function(purpose){
```
nikolai-b: instead of this we could do

```R
text_cycle_scenario <- function(purpose){
  gsub("baseline", "scenario", text_cycle_baseline)
}
```

regions_www/m/server.R
nikolai-b: Yes, might take me a while though.

regions_www/m/server.R
```diff
@@ -19,804 +19,1071 @@
...
+      } else {
+        line <- to_plot[[line_type]][to_plot[[line_type]]$id == id, ]
+        if (!is.null(line))
+          addPolylines(
+            leafletProxy("map"),
+            data = line,
+            color = "white",
+            opacity = 0.4,
+            layerId = "highlighted"
+          )
```
nikolai-b: Great, I'm assuming that most of the plotting code hasn't changed much (except checking if the data exists before trying to plot it).

regions_www/m/server.R
```diff
@@ -19,804 +19,1071 @@
...
-      addCircleMarkers(., data = to_plot$cents, radius = 0, group = "centres", opacity = 0.0) %>%
-      setView(., lng = to_plot$center_dim[1, 1], lat = to_plot$center_dim[1, 2], zoom = 10) %>%
+      # Centroids loaded invisibly to tell it the extent of the map
+      # Anna note: centroids do not exist for some purposes like schools - if ever want to *start* in schools layer need to change this to zones
+      # addCircleMarkers( .,
+      #   data = to_plot$centroids,
+      #   radius = 0,
+      #   group = "centroids",
+      #   opacity = 0.0
+      # ) %>%
```
nikolai-b: Is there a reason for this commented out code staying?

regions_www/m/server.R
```diff
@@ -19,804 +19,1071 @@
...
+
+  ## Adds map legend for zones
   observe({
-    input$map_base
+    input$purpose
+
+    # Define the legend title
+    if (input$purpose == "commute") { legend_title <- "% cycling to work"}
+    else if (input$purpose == "school") { legend_title <- "% cycling to school" }
+    else if (input$purpose == "alltrips") { legend_title <- "% trips cycled" }
```
nikolai-b: `switch`

regions_www/m/server.R
```diff
@@ -19,804 +19,1071 @@
...
-      content = function(file) { geojsonio::geojson_write(signif_sdf(to_plot$faster_route[codebook_r$`Variable name`]), file = file) }
-    )
+    # Create region name for files to be downloaded, so that they are accessible from the download_region.Rmd file
+    download_region <- data.frame(region_name = region$current)
+    saveRDS(download_region, file="../tabs/download_params_region_current.rds")
 
-    output$download_rq_geojson <- downloadHandler(
-      filename = function() { "routes_quiet.geojson"  },
-      content = function(file) { geojsonio::geojson_write(signif_sdf(to_plot$quieter_route[codebook_r$`Variable name`]), file = file) }
+    knitr::knit2html(quiet = T,
```
nikolai-b: We are `knitr` ing files on the fly?  
I'm unsure about this...
Why can't we do this so all the files have the html pre-generated

nikolai-b: We could have `repo_sha` in the file and gsub it.

I've not thought much about it but this worries me a bit.

regions_www/tabs/download_national.Rmd
```diff
@@ -0,0 +1,82 @@
+---
```
nikolai-b: Having `Rmd` files in a directory that was served by shiny server caused strange problems before, just one to watch if needs be.

regions_www/m/server.R
```diff
@@ -19,804 +19,1071 @@
...
+    region$current
 
-    output$download_z_geojson <- downloadHandler(
-      filename = function() { "zones.geojson"  },
-      content = function(file) { geojsonio::geojson_write(signif_sdf(to_plot$zones[codebook_z$`Variable name`]), file = file) }
-    )
+    region_stats_file <-
+      file.path("../tabs/region_stats", input$purpose, input$geography, region$current,
+                "region_stats.html")
+    if (file.exists(region_stats_file))
```
nikolai-b: Sorry if I've missed it but where are the `region_stats.html` files are generated.

regions_www/m/pct_shiny_funs.R
Robinlovelace: There's already a function that does this work.

regions_www/m/pct_shiny_funs.R
Robinlovelace: Agreed. Clear improvement, good refactor but it's not causing the website to crash so not a priority atm.

regions_www/m/pct_shiny_funs.R
Robinlovelace: :+1:

regions_www/m/server.R
Robinlovelace: Fair assumption.

regions_www/m/server.R
Robinlovelace: No.

regions_www/m/server.R
Robinlovelace: +1 good refactor.

regions_www/m/server.R
Robinlovelace: Also not sure..

regions_www/tabs/download_national.Rmd
Robinlovelace: Seems to work fine locally. Yes one to watch.

regions_www/m/server.R
Robinlovelace: Here: https://github.com/pctbike/pct-scripts/blob/master/07a_build_master.R

regions_www/m/pct_shiny_funs.R
nikolai-b: Not a priority but where else do we comment on things?

I'm going though the PR line by line and seeing what changes there are.

regions_www/m/server.R
nikolai-b: :cool: I'll delete it then

regions_www/m/server.R
nikolai-b: Great, thanks.

regions_www/tabs/download_national.Rmd
nikolai-b: Locally it will not be an issue as locally we don't run shiny server, we run shiny in R.

regions_www/m/server.R
nikolai-b: Ok, if this causes any issues I'd like to move this to static file generation please.

regions_www/m/pct_shiny_funs.R
nikolai-b: @Robinlovelace mind expanding on that 

> There's already a function that does this work.

Do you mean a function that creates the region names?  I can't find it, mind pointing it out?

nikolai-b:
@Robinlovelace & @usr110 RE: https://github.com/npct/pct-shiny/pull/596#issuecomment-325620252

I'm not against reverting this being merged, I'm also not suggesting it should be reverted.

Sorry I took over a week to review.  I believe our process is that large changes like this require sign off from more than one member of the tech team do we need to have that written down somewhere?

regions_www/m/pct_shiny_funs.R
Robinlovelace: Will do, on way to wedding though. Look for bbmisc in pct-scripts. U should find it. Can check Monday if not.

nikolai-b: Ah, in a different repo, thanks.  It should be the same in both places than.

regions_www/m/pct_shiny_funs.R
```diff
@@ -0,0 +1,758 @@
...
+#   "No. cyclists in Gender Equality"              = "gendereq_slc",
+#   "Change in deaths/year in Gender Equality"     = "gendereq_sideath_heat",
+#   "Change in CO2/year (kg) in Gender Equality"   = "gendereq_sico2",
+#   "No. cyclists in Go Dutch"                     = "dutch_slc",
+#   "Change in deaths/year in Go Dutch"            = "dutch_sideath_heat",
+#   "Change in CO2/year (kg) in Go Dutch"          = "dutch_sico2",
+#   "No. cyclists in Ebikes"                       = "ebike_slc",
+#   "Change in deaths/year in Ebikes"              = "ebike_sideath_heat",
+#   "Change in CO2/year (kg) in Ebikes"            = "ebike_sico2"
+# )
```
nikolai-b: Yay, let's remove this also then

regions_www/m/server.R
nikolai-b: Can we confirm this with @AnnaGoodman1 ?

regions_www/m/server.R
```diff
@@ -19,804 +19,1071 @@
...
-      filename = function() { "routes_quiet.geojson"  },
-      content = function(file) { geojsonio::geojson_write(signif_sdf(to_plot$quieter_route[codebook_r$`Variable name`]), file = file) }
+    knitr::knit2html(quiet = T,
+                     input = file.path("../tabs/download_region.Rmd"),
+                     output = file.path("../tabs/download_region_current.html"),
+                     envir = globalenv(), force_v1 = T
     )
+    # Re-read html to remove style
+    download_region_file <- file.path("../tabs/download_region_current.html")
+    download_region <- readLines(download_regio
```
nikolai-b: I'd love to remove the data tables, I spent a long time getting them to work and would be happy if we didn't have to maintain them, has this been agreed on then?

Robinlovelace: They could go if difficult to maintain but I think they are advantageous. Imagine if I don't have any computing skills but want to find the areas with the most car drivers. I could do this with a data table:

image

Worth checking with others. @AnnaGoodman1 @usr110 @JDWoodcock  what are your views?

regions_www/m/server.R
AnnaGoodman1: @nikolai-b , Right, sorry I hadn't appreciated that – feel free to revert it.  Also sorry for not looking at this pull request sooner, wwill now start replying below

regions_www/m/server.R
AnnaGoodman1: I see,  I think I couldn't work out when/how it would be used hence deleting it –  if you find it useful, of course it can be added back

regions_www/m/server.R
AnnaGoodman1: yes, that looks like it would be a much more elegant way to get to the same place

regions_www/m/server.R
AnnaGoodman1: oh I see, you mean one could create it in a temporary way and then feed it ddirectlyinto 'update_purposegeog' function without saving it? yes sure, if that works then great

regions_www/m/server.R
AnnaGoodman1: yes, agreed this is better

regions_www/m/server.R
AnnaGoodman1: yes, I think that's right. The only real change I can think of is defining bins by purpose, since so many  areas have >40% ttravel to school in the school layer - hence adding in one expert level for that:

      # Define bins and breaks (by purpose)
      if (input$purpose == "commute" || input$purpose=="alltrips") {
        zbins <- zbins_commute
        zbreaks <- zbreaks_commute
      } else if (input$purpose == "school") {
        zbins <- zbins_school
        zbreaks <- zbreaks_school
      }

regions_www/m/server.R
AnnaGoodman1: yes fine – it was there in our previous pct-shiny version and the comment "Centroids loaded invisibly to tell it the extent of the map" made it sound like it was doing something – but I agree it does not in fact seem to be needed

regions_www/m/server.R
AnnaGoodman1: switching to static would seem fine to me, and certainly if it causes any issues.  Could be moved to pct-scripts, like the region_stats files are

regions_www/tabs/download_national.Rmd
AnnaGoodman1: right, thanks. aanother reason to switch to static generation of the download pages, if this cause any issues

regions_www/m/pct_shiny_funs.R
AnnaGoodman1: yes, agreed we can!

regions_www/m/server.R
AnnaGoodman1: I think they should go, nice though they are in some ways.  The issue is that the data download pages  have now been considerably improved/extended in other ways, including now combining the lines and zones tabs to a single 'region data' page, and on that page showing all  available data downloads not just what is currently showing on the screen. 

I did think about it, and couldn't see anywhere to put these tables into the new data download pages without it being cluttered and confusing: esp since we would need both the zones and lines tables there on the same page.  The only good option I can see would be to have a show/hide-on-click function underneath the bullet points where we have e.g. region zone download data, saying 
"Click here to view an on-screen data table for selected information for the zones currently displayed on the map". then ditto for zones. but that sounds potentially difficult to build/maintain, plus having to create it for e.g. schools in due course. So my vote would be to remove this.

( Also even without many computing skills, downloading the relevant CSV 'all zones/ all flows' file  and doing a sort in that by the 'no cars' column should not be too hard?)

regions_www/m/pct_shiny_funs.R
AnnaGoodman1: yyes, agreed, that looks  much nicer!

regions_www/m/server.R
Robinlovelace: Sounds reasonable, cheers for the input: makes good sense to me. Another issue with them is that the zones do not have names that are meaningful to most people and cannot be displayed on the map this way so the advantage is pretty minimal.

regions_www/m/server.R
nikolai-b: Given the ∞ loop issue I'll leave this as it is!

PR #600, opened 2017-08-31
Wales

AnnaGoodman1:
Front page map to use IF we include Wales on main test server


Robinlovelace:
Great stuff.

AnnaGoodman1:
NB don't merge this yet! maybe we don't need it given Ali's fix to the link http://npct0.vs.mythic-beasts.com/shiny_interface/?r=wales

nikolai-b:
I think we should merge this as we already show it on the interface:
screenshot from 2017-09-04 20-26-03


Here is  on the test server:
screenshot from 2017-09-04 20-30-53

regions_www/index.html
```diff
@@ -62,9 +62,8 @@
           <p> Welcome to the Propensity to Cycle Tool (PCT) for England, which provides an evidence base to inform cycling investment.</p>
           <p>To run the PCT, first hover over a region on the map on the left to see the regional potential, then click on the region.</p>
           <p>For an introduction to the tool, its use, and the scenarios, please watch the video above.</p>
-          <p><strong>What's new :</strong> <a href="https://blog.pct.bike/2017/03/21/lower-level-super-output-area-route-network-now-up/", target = "_blank"> LSOA layer</a>;
-          <a href="http://www.cedar.iph.cam.ac.uk/resources/evidence/eb-14-englands-cycling-potential/", target = "_blank">Local-authority evidence briefing</a>
-            </p>
+          <p><strong>What's new :</strong> <a href="https://blog.pct.bike/2017/09/05/new-features/" target = "_blank">Cross-border flows, downloads, Wales & more</a>
```
nikolai-b: Really we should use `&amp;` instead of `&`. It works but I think its considered a bit safer as `&` is a keyword in html.

nikolai-b:
@AnnaGoodman1 I've made the this so it shows the Wales text for wales only.

AnnaGoodman1:
brilliant thanks, hopefully means ready to merge

AnnaGoodman1:
(From a quick test, the only bug I can see  is that if you click okay on the test banner, and then go to Wales, you don't get the Welsh banner unless you also press refresh. but I don't think this really matters, since in practice the Welsh banner is intended for the Live server. I would be happy for this to be merged)

regions_www/tabs/region_stats/commute/lsoa/wales/region_stats.html
```diff
@@ -46,11 +46,11 @@
 
 <p><br></p>
 
-<h3>The  Wales region</h3>
+<h3>Wales</h3>
 
 <p>Figure 1 shows a map of  Wales, including the regional boundary (thick black border), local authority boundaries (blue lines) and LSOA zones in the region (grey with a black border). The straight green lines represent the top 100 flows in terms of the number of commuters by any mode.</p>
 
-<p><em>Figure 1: Map of the  Wales region</em>
+<p><em>Figure 1: Map of Wales</em>
```
Robinlovelace: Thanks for adding these important changes - people from Wales would understandably want to be seen as more than a 'region'.

Robinlovelace:
Happy for this to be merged - will let you do this @AnnaGoodman1 as you did most of the work here. Suggest checking things are as expected on the test server directly after merging.

PR #603, opened 2017-09-04
Split to R and not R

AnnaGoodman1:


nikolai-b:
Yes, I'd like to add the script change to this so can we hold off merging.

PR #606, opened 2017-09-05
Use html

nikolai-b:
Hi,

Using knitr worries me and with lots of users as the html is built dynamically then if two users request a region at the same time one `download_params_region_current.rds` can be overwritten by another.

I've just converted them to HTML.


Robinlovelace:
Sounds good - less work to do on the server and as you say avoids rebuilding html on-the-fly. For once I am in favour of replacing .Rmds with .html! But where will the original .Rmd files that generate these live?

Robinlovelace:
Do you suggest we edit the raw html from now on?

AnnaGoodman1:
I think this is very good, thank you Nikolai. Robin, I think it would be fine to edit the raw HTML files from now on,, to me that seems fine and probably preferable

Robinlovelace:
OK so that's approved then - merging. Thanks Nikolai for the work and Anna for approving this.

PR #607, opened 2017-09-05
Revert "Add locizify"

nikolai-b:
I don't think we are producing a foreign language version so I'm removing locizify
This reverts commit 57c7fb3b8a6f485c92e01f3507a7b8da9d5e65d2.


AnnaGoodman1:
yes, we might do this one day but not at the moment, thanks

Robinlovelace:
Looks sensible to me, thanks for this. Plan to merge this afternoon giving ample chance for comments e.g. from @usr110 (cannot think of any unwanted side effects ;)

PR #610, opened 2017-09-05
Purpose region changing bugs

nikolai-b:


Robinlovelace:
Nice - not tested this but looks sensible, many thanks. As before, plan to merge today. @AnnaGoodman1 any comments before that welcome - could be good to have an example of the bug so we can reproduce the improvement @nikolai-b.

AnnaGoodman1:
this seems good to me, thanks!

PR #614, opened 2017-09-10
Fix loop

### LINKED ISSUE #613: Getting in a loop if changes made too fast
if you open one of the bigger regions, e.g. Wales or London,  and then immediately change both  middle layer to lower layer geography  plus change the scenario to e.g. go Dutch then it gets into a loop where it flicks between the census 2011 scenario and the new scenario. (See video in zip folder). This only happens, I think, in the big regions because in the smaller regions the app deals with the geography switch fast, before you have time to make the  scenario switch.  it happens both on internet edge and firefox. It only happens if you first switch geography (which is slow, as files need to reload), and then switch scenario (which is fast) – not the other way round

Ideally the app would handle making both changes simultaneously, or if not then at least force a revert to the current scenario – being a loop between the two seems the worst of all worlds... Would there be some way to force the app not to accept any input on the ui menu until the geography-switching was complete?

[video.zip](https://github.com/npct/pct-shiny/files/1283835/video.zip)

nikolai-b:
Fixes #613,
please test on schools and all trips.  Took me far longer to do than I'd have hoped.

R kept causing the update labels observe to be re-run even though `input$purpose` had not been changed (it is only changed in `update_purposegeog`) and `region$geography` had also not been changed (I put a print just before it).

I've put in a very hacky fix, sorry!

### DIFF
--- a/regions_www/m/server.R
+++ b/regions_www/m/server.R
@@ -217,6 +217,8 @@ shinyServer(function(input, output, session) {
   to_plot <- NULL
   helper <- NULL
   helper$e_lat_lng <- ""
+  helper$old_geog <- ""
+  helper$old_purpose <- ""
 
   ## Set  values of region
   observe({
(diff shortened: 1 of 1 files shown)


Robinlovelace:
Testing now, looks good to me.

AnnaGoodman1:
Thanks @nikolai-b , I agree this is a big improvement and I think now good enough to make live. @Robinlovelace , do you want to merge this? @nikolai-b , would you be able to then push the test server master branch to the live server as soon as convenient?

Robinlovelace:
Sure, just want to test this branch and try to break it first! Then plan to merge later today (by midday is the aim).

AnnaGoodman1:
great thanks. Given that James is waiting on sending this to Wales, I guess the sooner  the better...

Robinlovelace:
Tried hard to break it but cannot, even at LSOA layer for London: 

image

Robinlovelace:
Only issue I've found: it does not change automatically to the msoa level when on all-trips. This is not a major issue, so merging but one to think about - could you double check @AnnaGoodman1 and maybe open an issue if confirmed. see below for what I mean on this:

image

PR #615, opened 2017-09-10
Remove geojsonio pacakge

nikolai-b:
It is not used anymore, with the new R version it requires a few linux packages installed that I don't want to install on production.


Robinlovelace:
Indeed - we used to use `geojson_read()` as it was a little easier to use than `rgdal::readOGR()` but now the latter function guesses the driver based on the filename, as discussed with @usr110 that is no longer true. Great you're looking at updating the packages - worthwhile before the hackathon on 20th October.

PR #616, opened 2017-09-12
Add csv links to data download pages

AnnaGoodman1:


Robinlovelace:
Tested for Isle-of-Wight locally and seems to work fine: 

image

Pls double check on the test server for other regions now this is merged.

PR #619, opened 2017-09-12
update list geog if purpose changes

AnnaGoodman1:


Robinlovelace:
Tested: works.

Could not make it break - shifts to MSOA only on All tripls as illustrated below.

image

Issue #620, opened 2017-09-12
non-urgent issue - geographies not updating


AnnaGoodman1:
as @Robinlovelace points out, there is now a bug  with switching geographies:
1. Go to Manchester on the test server and select LSOA layer (purpose = commuting)
2. Switch the purpose to 'all trips'

On the screen, the zones correctly update to now be showing MSOAs.  However the geography is still shown as LSOA (see below), and LSOA is also given as an option on the drop-down menu when one firsts switches to all trips.  

I think I have found the problem, which is that the list of geographies needs to update if the purpose changes, and I fixed it here: https://github.com/npct/pct-shiny/pull/619 . @Robinlovelace / @Robinlovelace / @nikolai-b , could you test and merge if you are happy with this?

![untitled](https://user-images.githubusercontent.com/11818225/30322174-7d1be5c8-97b0-11e7-9fee-0d4fbb1757ad.png)

Robinlovelace:
Thanks for finding and fixing this Anna.

PR #622, opened 2017-09-12
Update production

Robinlovelace:
To all, especially @nikolai-b, there are a few minor bug fixes in here. Please review and merge when ready. Wider question: are there manual fixes we'll have to do each time we do a merge from test -> prod. Imagine it will be something we'll need to a few times over the coming months.


regions_www/m/server.R
```diff
@@ -333,9 +333,10 @@ shinyServer(function(input, output, session) {
    }
   }, priority = 3)
 
-  # Only requred to run if the region changes
+  # Only requred to run if the region changes (as that affects purpose) or the purpose changes (as that affects geographies)
   observe({
     region$current
+    input$purpose
```
nikolai-b: @AnnaGoodman1 I don't believe this should be needed

`region$purposes_present` and `region$geographies_present` only change when the `region$current` changes

https://github.com/npct/pct-shiny/blob/3374d9bb0b4a6b707431c19606a845ea48309aa6/regions_www/m/server.R#L284-L286

Would you mind explaining what this fixes?

Robinlovelace: This (before it did not switch back to MSOAs):

peek 2017-09-13 07-39

AnnaGoodman1: @nikolai-b : fixes this issue: https://github.com/npct/pct-shiny/issues/620 (sorry, should have cross-referenced)

nikolai-b: Right, that makes sense, thanks

Robinlovelace:
@nikolai-b planning to merge this by midday.

PR #624, opened 2017-09-12
Fix stats table missing styling

### LINKED ISSUE #618: minor issue - table formatting on region stats files is lost
sorry, I realise that along the way I have lost the table formatting on the region stats (formerly model output) tab. @usr110, I think it was you who did this originally?  see below for the former formatting and the current formatting.  I copied the code I could see, such that each table is defined as having table class='region_stats_table'>, but I now realise that the attributes of 'region_stats_table' must have been defined somewhere that I did not find (  they did not seem to be defined in set-up.R in the old pct-load, now called pct-scripts, nor in the Rmd file making model_output).

it is not very important, but if Ali/anyone can remember where/how the table class properties were defined, and reinstate it at some point,  that would make the tables look a bit nicer.

old formatting
zz

new formatting
xx

nikolai-b:
Fixes #618

Looks like the table class got renamed, update the CSS

### DIFF
--- a/regions_www/www/stylesheet.css
+++ b/regions_www/www/stylesheet.css
@@ -102,18 +102,16 @@ Add CSS for tables. Taken from:
 https://onlinecourses.science.psu.edu/statprogram/markdown
 */
 
-.model_output_table th {
+.region_stats_table th {
   background-color:#E0E0E0 ;
   border-bottom:1px solid black;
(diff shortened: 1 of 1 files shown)


Robinlovelace:
Great fix - tested and works as illustrated below:

image

Issue #626, opened 2017-09-13
where are the centroids?


AnnaGoodman1:
when I run the app locally then I see the centroids, but they are not there on either the test server or the live server, at the equivalent zoom level (see below).  Not sure why this could be – assuming the code is identical, my only thought is could the c.Rds files in pct-outputs-regional-R somehow not have been copied across to the server?
![untitled](https://user-images.githubusercontent.com/11818225/30367596-8f6904f2-9866-11e7-96af-6267ce5bad2a.png)

@Robinlovelace / @nikolai-b , not sure if either of you could check this?

Robinlovelace:
Well spotted - I confirm they are not there.

Robinlovelace:
Update: they are not there on my local instance either.

Robinlovelace:
Do you have the latest version of the repo? I'm on the master branch.

AnnaGoodman1:
thanks Robin: yes I'm on the master branch in pct-shiny, having updated it, and I do still see them in my local run.  When you have time, you could perhaps check if the c.Rds are present in your local copy of pct-outputs-regional-R (they should be : see e.g. https://github.com/npct/pct-outputs-regional-R/blob/master/commute/msoa/avon/c.Rds).  If they are, but it's still not working locally few, then that seems strange indeed...

AnnaGoodman1:
Okay sorry, I was wrong, I was still seeing the centroids locally after updating pct-shiny... But now don't. so maybe one of them more recent changes, I will see if I can work out what is going on

AnnaGoodman1:
sorry, I think what I said before about being able to see the centroids locally was just a mistake – I could only see them because I was printing 'input$map_zoom' as a test (but one that fixed the problem!).  So actually I think this problem may have been therefore while, probably introduced by me.  This maybe fixes it, splitting up the zones and centroids 'observe' commands: https://github.com/npct/pct-shiny/pull/627 . As I check it it seems to work, but would be grateful if @nikolai-b / @Robinlovelace could also look and confirm they are happy that this won't break something else I haven't thought of

(NB the reason one can't just observe 'input$map_zoom' in the original combined zone/centroid function is this issue https://github.com/npct/pct-shiny/issues/597 of zones refreshing every time you change the zoom level)

AnnaGoodman1:
[...]

PR #627, opened 2017-09-13
attempt to restore centroids

AnnaGoodman1:
I think this may work as a fix for https://github.com/npct/pct-shiny/issues/626 , but please check! I can't see anything else breaking, but a paranoid it might...

### DIFF
--- a/regions_www/m/server.R
+++ b/regions_www/m/server.R
@@ -551,17 +551,16 @@ shinyServer(function(input, output, session) {
   })
 
 
-  ## Displays zones and centroids
+  ## Displays zones
   observe({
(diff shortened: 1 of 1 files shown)


Robinlovelace:
This looks good to me and I could not find any un-intended consequences when I tested the changes locally so I'm happy for this to be merged. I'd like @nikolai-b to comment on this before it is merged though to check there's not a simpler solution and ensure that it doesn't have any other consequences that I'm not aware of. Looking at the code it looks like a pretty complex piece of programming. Great work!

nikolai-b:
Great, that makes sense!

PR #628, opened 2017-09-13
Fix sort by heat

AnnaGoodman1:
see https://github.com/npct/pct-shiny/issues/629 for details

### DIFF
--- a/regions_www/m/server.R
+++ b/regions_www/m/server.R
@@ -417,8 +417,13 @@ shinyServer(function(input, output, session) {
       if (all(!keep))
         return(NULL)
       lines_in_bb <- lines[drop(keep),]
-      # Sort by the absolute values
-      lines_in_bb[tail(order(abs(lines_in_bb[[line_data()]])), nos),]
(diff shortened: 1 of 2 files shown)


regions_www/m/server.R
```diff
@@ -417,8 +417,13 @@ shinyServer(function(input, output, session) {
       if (all(!keep))
         return(NULL)
       lines_in_bb <- lines[drop(keep),]
-      # Sort by the absolute values
-      lines_in_bb[tail(order(abs(lines_in_bb[[line_data()]])), nos),]
+      # Sort low-to-high for reduction in deaths (can't use absolute values as no. deaths can be a positive number, i.e. health disbenefit)
+      if (grepl(c("sideath_heat"), line_data())) {
```
nikolai-b: We don't let people search for `sivalue_heat` but that can also be -ve or +ve, just checking we've no intention of adding that.  If not than :+1: from me, really nice spot.

Robinlovelace:
Looks great. Aim to test later tonight and merge unless I find anything untoward.

nikolai-b:
I've tested so I'm happy to merge and we can make an create an issue if we find one later

PR #630, opened 2017-09-13
Minor style

nikolai-b:


regions_www/m/server.R
```diff
@@ -253,64 +261,51 @@ shinyServer(function(input, output, session) {
...
+             }
+           },
+           "school"= { region_geo_change_to <- "lsoa" },
+           "alltrips"= { region_geo_change_to <- "msoa" }
+    )
+
     # Only trigger geography changes if required.
-    if (is.na(region$geography) || region_geo_change_to != region$geography){
-      region$geography <<- region_geo_change_to
+    if (!identical(region_geo_change_to, region$geography)) {
```
Robinlovelace: neater and deals with NAs - awesome!

regions_www/m/server.R
```diff
@@ -253,64 +261,51 @@ shinyServer(function(input, output, session) {
...
+      region$geography <- region_geo_change_to
     }
 
     # Set data_dir
-    region$data_dir <<- file.path(data_regional_root, input_purpose(), region$geography, region$current)
+    region$data_dir <- file.path(data_regional_root, input_purpose(), region$geography, region$current)
```
Robinlovelace: Was always wondering why you used global assignment..

regions_www/m/server.R
```diff
@@ -253,64 +261,51 @@ shinyServer(function(input, output, session) {
...
 
     # Identify that region repopulation has happened
-    region$repopulate_region <<- T
+    region$repopulate_region <- T
```
Robinlovelace: From a style perspective `TRUE` is better.

regions_www/m/server.R
```diff
@@ -253,64 +261,51 @@ shinyServer(function(input, output, session) {
...
-      to_plot$straight_lines <<- readRDS(file.path(region$data_dir, "l.Rds"))
-    } else {
-      to_plot$straight_lines <<- NULL
-    }
-
-    if (file.exists(file.path(region$data_dir, "rf.Rds"))) {
-      to_plot$routes_fast <<- readRDS(file.path(region$data_dir, "rf.Rds"))
-    } else {
-      to_plot$routes_fast <<- NULL
+    to_plot$zones <<- load_data(file.path(region$data_dir, "z.Rds"))
```
Robinlovelace: Why `<<-` here but not previously? Suggest `<-` unless it breaks something for consistency.

AnnaGoodman1:
apart from the one specific comment about get_pretty_region_names, this looks fine to me (as far as I can tell - some of the changes are a little beyond me, e.g. I have not yet got to grips with global assignment).thanks @nikolai-b

regions_www/m/pct_shiny_funs.R
```diff
@@ -3,15 +3,16 @@
...
-get_pretty_region_name <- function(region_name, the = T){
-  if (the == T) {
-    region_name <- gsub("isle-of-wight", "the-isle-of-wight", region_name, perl=TRUE)
-    region_name <- gsub("north-east", "the-north-east", region_name, perl=TRUE)
-    region_name <- gsub("west-midlands", "the-west-midlands", region_name, perl=TRUE)
-  }
-  region_name <- gsub("(^|-)([[:alpha:]])", " \\U\\2", region_name, perl=TRUE)
+get_pretty_region_name <- function(region_name_in, the = T){
+  regions_thes <- c(
+    "isle-of-wight" = "The Isle of Wight",
```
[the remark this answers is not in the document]
nikolai-b: Sure, I've re-added the "The" ability, I was only doing the change for this repo and hadn't thought about modifying `pct-script`.


Before this change
```R
get_pretty_region_name("isle-of-wight")
=> " the Isle of Wight"
```
(lower case "the" and an unwanted space at the start)

With this change:
```R
get_pretty_region_name("isle-of-wight")
=> "The Isle of Wight"
```
so I don't really follow about part about capitalisation. Would you mind giving an example of how this isn't working?

It still works for regions like:
```R
get_pretty_region_name("cornwall-and-isles-of-scilly")
=> "Cornwall and Isles of Scilly"
``

regions_www/m/server.R
nikolai-b: Sadly not.  `to_plot` is not a `reactiveValue` and so global assignment is required.
We can change it but not in this PR please.

regions_www/m/server.R
nikolai-b: When I added them it was `<-` https://github.com/npct/pct-shiny/commit/11e1d90ac456bbe34204eb0ffd0b042538ee3648 but it must have been changed along the way :woman_shrugging:

Robinlovelace:
Ready for this to be merged @nikolai-b ? Happy for this to be merged @AnnaGoodman1 ?

Those edge-case names are a little annoying, think Nikolai's solution is a little neater and would be a minor switch-up in pct-scripts.

regions_www/m/pct_shiny_funs.R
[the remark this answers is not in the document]
AnnaGoodman1: @nikolai-b , the issue with capitalisation is that I think  
get_pretty_region_name("isle-of-wight") => "the Isle of Wight" 
was correct and 
get_pretty_region_name("isle-of-wight") => "The Isle of Wight"
is wrong.  The 'the' on Isle of Wight (or the other 2 edge regions) is not routinely capitalised (see e.g. https://en.wikipedia.org/wiki/Isle_of_Wight).  So your change would turn the current text "This page contains data downloads for the Isle of Wight" (right) to "This page contains data downloads for The Isle of Wight" (not right).  sorry, is a picky point... [we deliberately never start a sentence with the region, so it can be always lower case]

I agree that removing the unwanted space is better, thanks. so perhaps just modify the capital 'The' to small 'the' in lines 8-10? then I am happy :) and I will update pct-scripts accordingly

...thanks for adding the 'the' switch function

AnnaGoodman1:
@Robinlovelace : am *almost* happy, just one remaining thing to sort out on the region name capialisation issue (See comment above).

Robinlovelace:
Yes I see your point about the (not The!) Isle of Wight. When that's resolved it can be merged. Great to have many eyes on this. Priority is to get these excellent recent changes onto production - plan to do that by midday today.

regions_www/m/pct_shiny_funs.R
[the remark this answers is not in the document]
nikolai-b: Right, I understand now.  Thanks.

nikolai-b:
Fine to be picky, that is what the PR review process is for!

Issue #621, opened 2017-09-12
csv download bug in firefox


AnnaGoodman1:
in the course of adding csv downloads the zones, I notice that the CSV download function was already not working well for the existing 'all flows' csv files (or the new csv zones files) on Firefox browsers. Microsoft edge and Google Chrome give one a CSV download if you click the link, but Firefox opens the CSV in a new page
![untitled](https://user-images.githubusercontent.com/11818225/30325723-b07299fa-97bd-11e7-9d7d-56f9cec308de.png)

this does not happen for the codebooks because they are using cdn.rawgit, just the lfs files... And one can then right click the Firefox page to "save page", at which point it saves as csv.  But obviously nicer if it did this all in one go

Any ideas on this, anyone?

Robinlovelace:
I don't think this is a bug. It's a browser issue. If you press `Ctl-S` it should save it as a .csv file. I think mentioning that in the manual should be sufficient to close this issue.

AnnaGoodman1:
I agree that is def an option, but I guess will still annoy/puzzle some users. Lets see if @nikolai-b / @usr110 have any thoughts on how to make firefox handle this download differently - if not then I agree we can live with it.

nikolai-b:
This is a bug!

The browser deals with different content based on the https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type and git hub (purposely) doesn't set the content type header.

nikolai-b:
Sadly I think the answer would be to use S3 as github isn't the correct place to serve files.

AnnaGoodman1:
Thanks @nikolai-b... that is a bit of a pain, but perhaps  (not urgent) is worth transitioning to S3 in that case. in which case, sorry, I was wrong to talk you out of it before... 

if we did this then I would suggest that pct-outputs-national and pct-outputs-regional-notR would be deleted from git and moved to Amazon S3.  (ideally, for consistency, we might move pct-outputs-regional-R to S3 too - but I think perhaps that needs to be on git for cloning to the server?).  these repos are in any case much too big to readily clone on git.  Would that seem okay to you @Robinlovelace ? 

or the alternative is that we leave it as it is and live with it – we could ask James and Rachel how much of a problem they think it is at the next team meeting

Robinlovelace:
[...]

PR #631, opened 2017-09-14
temporary fix for https://github.com/npct/pct-shiny/issues/621

AnnaGoodman1:
warning users about the 'csv download of firefox' issue

### DIFF
--- a/non_www/tabs/download_national.html
+++ b/non_www/tabs/download_national.html
@@ -10,7 +10,7 @@
       <div class="col-md-12 text-left" style="padding-top:10px;">
         <h3>National Data</h3>
         <p>This page contains data downloads for the whole of England. These are available as CSV files containing the zone and line baseline and scenario data, and in geographic files that can be read by R (.Rds) or GIS programs such as QGIS (.geojson).  Very large geographic files are available in R format only.</p>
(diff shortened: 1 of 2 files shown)


nikolai-b:
Up to you but we could also use `$.browser.mozilla` to display the warning for FF users only?

Robinlovelace:
I think it's good to warn all users, so they know if they try on a different computer or with someone who uses FF not to be afraid.

AnnaGoodman1:
@nikolai-b , if it is easy to make that sentence firefox specific then to me that would seem a bit better - reduces the amount of text non FF people have to read...

Robinlovelace:
Everyone should see the same text in the download tabs. I feel pretty strongly about that (as strongly as you can for such a minor issue ; ).

Robinlovelace:
I've checked the text and I think it's a clear improvement that will benefit all users. The mention of filenames also minimises ambiguity. This has no side-effects so merging. Can think about browser specific messages later down the line if there's demand or if people feel strongly that the html should vary from computer to computer (which I cannot imagine).

nikolai-b:
I disagree, I think as different browsers do different things and as such we should only show the warning when relevant but as you feel strongly (and I don't) then we can leave it as it is.

PR #640, opened 2017-11-02
Add loading spinner

### LINKED ISSUE #639: Communicate delay
It is not clear to users when the server is doing something as with large regions there can be a delay after an region is changed or a button is clicked.

nikolai-b:
Needs feedback and tweaking but something like has been asked for a while.
spinner

Fixes #639

### DIFF
--- a/regions_www/m/server.R
+++ b/regions_www/m/server.R
@@ -239,6 +239,7 @@ shinyServer(function(input, output, session) {
 
   ## Set  values of region
   observe({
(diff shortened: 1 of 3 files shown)


Robinlovelace:
Tested. Works (see below). Great work!

@AnnaGoodman1 and @usr110 would you like to test before I merge this? Plan to merge early next week.

peek 2017-11-03 08-30

PR #644, opened 2017-11-05
Change page title when region changes

### LINKED ISSUE #636: Title (for URL bar / search results) does not state area
http://pct.bike/m/?r=isle-of-wight

has

`<title>Propensity to Cycle Tool</title>`

but it would be more useful, for back/forward navigation, for this to be

`<title>Propensity to Cycle Tool - Isle of Wight</title>`

nikolai-b:
Fixes #636


Robinlovelace:
Tested. Works. Minor improvement to UI with no side effects so merging.

PR #650, opened 2018-01-29
Add Welsh LSOA tiles

### LINKED ISSUE #647: Do tiles for Wales
@nikolai-b: I have created a release of the England plus Wales raster .tif files here: https://github.com/npct/pct-inputs/releases/edit/untagged-c6cfec2972276d2c500b 

There are 15 files, for the five scenarios * 3 different resolutions for different levels of zoom (e.g. ‘-50’ is a cell size of 50m2. The ‘-10’ version is also available regional/national download files for users).  If I understood correctly from Malcolm these are what you will need to create the tiles that are served for the LSOA raster layer, but let me know if you need anything else or I can provide them in a more helpful format. 

I think the code you will use for this has now been archived here: https://github.com/npct-archive/pct-lsoa-vis - but actually (sorry) it should probably be unarchived and moved back to npct, since it is still in use? Or if only the files in pct-lsoa-vis/generate/ are needed and not those in pct-lsoa-vis/zones/ then perhaps it could be added as an additional step in npct/pct-scripts, e.g. as files prefixed with 10a_, 10b_, 10c_ ?

Obviously good if you can keep a copy of the current England-only tiles somewhere, in case anything goes wrong and we need to avert.  Otherwise it will be fine to switch to having the England+Wales tiles being what the server is deploying as and when those are available.  I have done some checking of what I created, so hope it is all okay, but can do some more checking when the tiles are up on the test server

nikolai-b:
On a subdirectory on the tileserver
Closes #647

### DIFF
--- a/regions_www/m/server.R
+++ b/regions_www/m/server.R
@@ -763,7 +763,7 @@ shinyServer(function(input, output, session) {
     # region$repopulate_region
     shinyjs::showElement(id = "loading")
     if (input$line_type == "lsoa_base_map") {
-      urlTemplate <- paste0("https://npttile.vs.mythic-beasts.com/",input$scenario,"/{z}/{x}/{y}.png")
+      urlTemplate <- paste0("https://npttile.vs.mythic-beasts.com/en-cy/",input$scenario,"/{z}/{x}/{y}.png")
 
       leafletProxy("map") %>%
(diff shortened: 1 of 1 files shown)


Robinlovelace:
Tested, works, great work: 

image

PR #651, opened 2018-01-31
Pct schools jan18 rebased

nikolai-b:
From #648 
Hi @AnnaGoodman1 I redid you commits so we can remove the old tabs without an issue.  I also tweaked the rounding functions and the font colour.  I don't think I've got the schools data but if your happy then I am.

Sorry to be a pain but it is much easier to clean up code when its touched or next time I am likely to add a bug.


regions_www/m/pct_shiny_funs.R
```diff
@@ -43,18 +43,10 @@ round_dp <- function(expression, small_threshold = 0.05, large_threshold = 0.5){
...
- value <- round(100 * expression)
-  for(i in 1:length(data)) {
-    if(!is.na(expression[i]) && !is.nan(expression[i]) && expression[i]>0 && expression[i]<0.005) {
-      value[i] <- round(100 * expression[i], 1)
-    }
-    if (is.nan(value[i]) || is.na(value[i])) { value[i] <- "-" }
-  }
- value
+round_percent <- function(expression){
+  round_dp(expression * 100, small_threshold = 0.0, large_threshold = 0.005)
```
AnnaGoodman1: I think the large threshold value needs to be 0.5 not 0.005, since I think the *100 comes first and then the threshold is applied? Unless I make this change, the percentage shown below appears as 0% not 0.5%

untitled

regions_www/m/pct_shiny_funs.R
```diff
@@ -164,9 +164,9 @@ data_filter <- function(scenario, type){
   ifelse(scenario == "olc", "bicycle", paste(scenario, type, sep = "_"))
 }
 
-# Create a new variable called font_colour which changes into red colour when change in death/yr is negative
-positive_red <- function(dataset, scenario, type){
-  dataset@data$font_colour <- ifelse(round(dataset[[data_filter(scenario, type)]]) <0, "red", "black")
+# Return red if the data is positive
```
AnnaGoodman1: yes, nice, except that it is actually red if the data is negative not positive – I have renamed function accordingly in one of my commits

AnnaGoodman1:
hi @nikolai-b , great thanks very much for this – and totally agree it is nice to clean things up as we go.

 I think I might have found two small issues, see my commits/review comments.  Otherwise I am happy to sign off on this and have it merged (and the non-rebased pct-schools branch deleted).  [My R is not really good enough to be very sure that my review would have picked up any issues, but I have tried to test the changes in the interface and they all seem fine.]

If you want to try out some schools data yourself, then you can download the files here: https://github.com/npct/pct-shiny/releases/tag/v1.0 - if you follow the instructions about putting them in a folder in the correct place, then you should be able to get Isle of Wight to run

Robinlovelace:
Not had a chance to test this - @nikolai-b if you and @AnnaGoodman1 are confident this is good and have tested it I'm more than happy for it to be merged. Many thanks for the mid-project tidy!

nikolai-b:
Ah, yes `< 0` is negative and thanks for the percentage rounding spot @AnnaGoodman1.

PR #652, opened 2018-02-08
School layer updates 180206

AnnaGoodman1:
Changing what data is displayed on the schools layer, to  comply with data protection


AnnaGoodman1:
@Robinlovelace , these changes and parallel ones in pct-scripts are in line with methods set out in this work doc (also at Dropbox\PCT\2_WorkInProgress\Anna\180207_ApplyingSDCtoschools.doc if you want to edit).  I think they meet all the NPD rules, so if you are happy with the principle you can perhaps check the code looks okay to you and then we may be able to merge.
[180207_ApplyingSDCtoschools.docx](https://github.com/npct/pct-shiny/files/1708057/180207_ApplyingSDCtoschools.docx)

@nikolai-b ,  any comments also welcome.  In particular, could you please check that this commit does not look as if it will break anything to you? (https://github.com/npct/pct-shiny/pull/652/commits/7b88d1acef67ee8cafff3ba65933be84c7f05e6a). For now it does not actually do anything useful just tries to redirect where a file is saved, but seems more sensible structure for saving the front page scenario data that will be available for different layers, so that can be introduced in due course.

Also, for both @Robinlovelace  and @nikolai-b , To let you test the system I have uploaded to pct-regional-R the Isle of Wight files that go with the layer 'school' and 'school_public'.  Basically, the files in 'school' are what our scripts read to create the layer in a way that looks nice and does not have lots of blank areas of the map and should become private; and the files in 'school_public' are the ones that people can download. So, before this can be rolled out more properly, we would need to find a way to stop most people being able to see into 'school'.  This could either be password protecting those folders (e.g.  E.g. https://gist.github.com/darryl-snow/3817253 ??) or else bypassing git for these private datasets altogether, and just somehow putting them on the server directly? or maybe something else, I don't really know on this one

regions_www/m/pct_shiny_funs.R
```diff
@@ -165,10 +170,27 @@ data_filter <- function(scenario, type){
...
 
 # Return red if the data is negative
-negative_red <- function(dataset, scenario, type, round_digits = 0){
-  ifelse(round(dataset[[data_filter(scenario, type)]], round_digits) < 0, "red", "inherit")
+negative_red <- function(data, scenario, type, round_digits = 0){
+  ifelse(round(data[[data_filter(scenario, type)]], round_digits) < 0, "red", "inherit")
 }
 
+# Identify small cells in schools layer
+school_smallcell <- function(expression, return_tf = F, d = F){
```
Robinlovelace: Neat, looks good. Can be rewritten more concisely as:

```r
school_smallcell2 <- function(x, return_tf = F, d = F){
  if(!return_tf) {
    if(d) res = case_when(x %in% 1:5 ~ "1 to 5", TRUE ~ as.character(x)) else
      res = case_when(x %in% 1:2 ~ "1 to 2", TRUE ~ as.character(x))
  } else {
    if(d) res = case_when(x %in% 1:5 ~ TRUE, TRUE ~ FALSE) else
      res = case_when(x %in% 1:2 ~ TRUE, TRUE ~ FALSE)
  }
  res
}
```

Here's a [reprex](https://www.tidyverse.org/help/) testing both, your one is fine, was just looking if it could be made neater but yours is more explicit which is good:

``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
x = 1:6

# original way
school_smallcell <- function(expression, return_tf = F, d = F) {
  if (d) {
    smallcell <- (expression > 0 & expression <= 5)  # Upper bound to suppress cells if 5 in schools [d], otherwise 2 [z, rnet]
    expression[smallcell] <- "1 to 5"
  } else {
    smallcell <- (expression > 0 & expression <= 2)
    expression[smallcell] <- "1 or 2"
  }
  if (return_tf) {
    smallcell
  } else {
    expression
  }
}

# recode way
school_smallcell2 <- function(x, return_tf = F, d = F) {
  if (!return_tf) {
    if (d) 
      res = case_when(x %in% 1:5 ~ "1 to 5", TRUE ~ as.character(x)) else res = case_when(x %in% 1:2 ~ "1 to 2", TRUE ~ as.character(x))
  } else {
    if (d) 
      res = case_when(x %in% 1:5 ~ TRUE, TRUE ~ FALSE) else res = case_when(x %in% 1:2 ~ TRUE, TRUE ~ FALSE)
  }
  res
}
school_smallcell(x)
#> [1] "1 or 2" "1 or 2" "3"      "4"      "5"      "6"
school_smallcell(x, d = TRUE)
#> [1] "1 to 5" "1 to 5" "1 to 5" "1 to 5" "1 to 5" "6"
school_smallcell(x, return_tf = TRUE)
#> [1]  TRUE  TRUE FALSE FALSE FALSE FALSE
school_smallcell2(x)
#> [1] "1 to 2" "1 to 2" "3"      "4"      "5"      "6"
school_smallcell2(x, d = TRUE)
#> [1] "1 to 5" "1 to 5" "1 to 5" "1 to 5" "1 to 5" "6"
school_smallcell2(x, return_tf = TRUE)
#> [1]  TRUE  TRUE FALSE FALSE FALSE FALSE
```

Robinlovelace:
Great set of changes Anna, many thanks. There are many ways to transfer files securely, will let @nikolai-b comment on the most appropriate for getting it on the server. Overall these changes look good and I don't have comments (other than the stylistic one which has no impact on the results and does not need to lead to a change).

've not had a chance to test them - would you mind giving this a spin Nikolai to see if it works and if you foresee any issues with deployment?

non_www/tabs/school/download_national.html
```diff
@@ -9,33 +9,34 @@
...
+            <a href="https://github.com/npct/pct-outputs-national/raw/master/school_public//lad_attributes.csv" target="_blank" download="lad_attributes.csv" onclick="ga('send', 'event', 'download', 'lad_attributes.csv');">
               csv
             </a>
-            <a href="https://github.com/npct/pct-outputs-national/raw/master/school//lad.Rds" target="_blank" download="lad.Rds" onclick="ga('send', 'event', 'download', 'lad.Rds');">
+            <a href="https://github.com/npct/pct-outputs-national/raw/master/school_public//lad.Rds" target="_blank" download="lad.Rds" onclick="ga('send', 'event', 'download', 'lad.Rds');">
               Rds
             </a>
-            <a href="https://github.com/npct/pct-outputs-national/raw/master/school//lad.geojson" target="_blank" download="lad.geojson" onclick="ga('send', 'event', 'download', 'lad.geojson');">
+            <a href="https://github.com/npct/pct-outputs-national/raw/master/school_public//lad.geojson" target="_blank" download="lad.geojson" onclick="ga('send', 'event', 'download', 'lad.geojson');">
```
nikolai-b: Very minor, I missed it last time but the URLs above have two `//`, i.e.
```html
school_public//lad.geojson
```
it shouldn't effect anything but might on some strange browser.

also, I'm assuming that we are going to put the schools data in `school_public` in the repos?

non_www/tabs/school/download_national.html
```diff
@@ -9,33 +9,34 @@
...
       </div>
 
       <div class="col-md-12 text-left" style="padding-top:10px;">
         <h4><a name="natschoolzones"></a>School travel data for zones</h4>
-        <p>Download anonymised National School Census 2011 baseline data and scenario attribute data for all zones in England. These files present aggregate data for all children <b>living in a given zone</b>. </p>
+        <p>Download anonymised National School Census 2011 baseline data and scenario attribute data for all zones in England. These files present aggregate data for all children <b>living in a given zone</b></p>
+        <p>In the LSOA downloads, baseline and scenario data are not presented for cells containing 1 or 2 children.  Instead these data are set as missing ('NA').  These data are, however, included in the local authority downloads.</p>
```
nikolai-b: Nice, do we only have `NA` in these circumstances, i.e. do we have 0 if there are no children (or schools)?

non_www/tabs/school/download_national.html
```diff
@@ -49,15 +50,17 @@ <h4><a name="natschoolzones"></a>School travel data for zones</h4>
       <div class="col-md-12 text-left" style="padding-top:10px;">
         <h4><a name="natschooldestinations"></a>School travel data for schools</h4>
```
nikolai-b: Minor again but this is a bit of a clunky heading
#### School travel data for schools

Can't it just be
#### School travel data

non_www/tabs/school/download_national.html
nikolai-b: The same `//` exists in https://github.com/npct/pct-shiny/blob/master/non_www/tabs/commute/download_national.html which is probably my fault. Sorry!

nikolai-b: Might also be nice to separate these Google Analytics events out.  We could use:
```
onclick="ga('send', 'event', 'download', 'school_public/lad.geojson')
```

and same with all the other events.  I'm happy to make that change if you like.

non_www/tabs/school/download_region.html
```diff
@@ -61,13 +63,14 @@ <h4><a name="regschooldestinations"></a>School travel data for schools</h4>
      <div class="col-md-12 text-left" style="padding-top:10px;">
         <h4><a name="regschoolroutenetSchool"></a>School travel data for the route network</h4>
         <p>Download anonymised National School Census 2011 baseline and scenario cycling estimates for the route network in <!--region_name-->.  This was calculated using within-region flows (i.e. children who live and attend school in <!--region_name-->) containing at least 5 children.</p>
+        <p>Baseline and scenario data are not presented for route segments containing 1 or 2 cyclists.  Instead these data are set as missing ('NA').</p>
```
nikolai-b: Wow, the requirements are very strange.  So it lets us show that there is 1 or 2 but not say which?  Thanks for digging into this @AnnaGoodman1 !

regions_www/m/pct_shiny_funs.R
```diff
@@ -30,11 +30,11 @@ normalise <- function(values, min = 0, max = 1){
...
 
-## Create percentages such that rounded to XDP if >0% and <0.5%.  [NB can be NaN if denominator zero]
+## Create Values such that rounded to XDP if >0 and <0.5.  [NB can be NaN if denominator zero]
 round_dp <- function(expression, small_threshold = 0.05, large_threshold = 0.5){
-  small <- is.finite(expression) & expression > 0 & expression < small_threshold
-  medium <- is.finite(expression) & expression >= small_threshold & expression < large_threshold
-  large <- is.finite(expression) & expression >= large_threshold
+  small <- is.finite(expression) & expression > (small_threshold*-1) & expression < small_threshold
+  medium <- is.finite(expression) & ((expression <= (small_threshold*-1) & expression > (large_threshold*-1)) | (expression >= small_threshold & expression < large_threshold))
+  large <- is.finite(expression) & ((expression <= (large_threshold*-1)) | (expression >= large_threshold))
```
nikolai-b: :clap: nice. 
I don't think you need to `expression > (small_threshold*-1)`, you can just do `expression > -small_threshold`

regions_www/tabs/region_stats/school/lsoa/avon/region_stats.html
```diff
@@ -63,7 +65,7 @@ <h3>Cycling levels in each scenario for Avon</h3>
 
 <p>Table 1 summarises the proportion of children cycling, walking, driving and using other modes under each scenario in Avon, based on aggregating across all residents in all the LSOA zones.</p>
 
-<p><em>Table 1: Travel to school mode share in the NSC 2011 and in scenarios among 141,727 children living in Avon</em> </p>
+<p><em>Table 1: Travel to school mode share in the NSC 2011 and in scenarios among 141,726 children living in Avon</em> </p>
```
nikolai-b: We lost one!

nikolai-b:
Hmm confidential data.  There are many options but I'd like to know the Census guidelines / requirements first.  I had a look but I couldn't find them, any idea where they are?  My preference would be to manipulate the data before hand so instead of the 'NA' is just had 2 (or 5) in which case we don't really need to make it private if I understand correctly.

regions_www/m/pct_shiny_funs.R
AnnaGoodman1: oh nice... happy for you to change it if you would like, I wasn't sure if on balance you thought conciseness or explicitness was better

non_www/tabs/school/download_national.html
AnnaGoodman1: I think the // is a legacy from when we used a function to generate those URLs, and the lad. files are one level up from the other ones. I have now replaced this.

good idea in terms of Google analytics – I have made the change I think you mean

non_www/tabs/school/download_national.html
AnnaGoodman1: yes, have a look e.g. here: https://github.com/npct/pct-outputs-regional-notR/blob/master/school_public/lsoa/isle-of-wight/z_attributes.csv 
if there are no children then we do have 0, since this is allowed by the NPD SDC rules

non_www/tabs/school/download_national.html
AnnaGoodman1: sigh, yes I know – the problem was I was trying to separate out the type of travel (school travel versus commuting) with the unit of presentation (zones versus schools).  but I agree this is not very good, I have changed it to "travel data for zones" and "travel data for schools'

non_www/tabs/school/download_region.html
AnnaGoodman1: that is my understanding, see page 36 of the NPD user guide:

"NPD figure(s) at a local level must be based on *no fewer than 3 cases* or else suppressed (e.g. replaced with an X, with a note “* in this table means the figures have been suppressed due to small numbers”. *Zero is allowed*)"

https://www.gov.uk/government/uploads/system/uploads/attachment_data/file/472700/NPD_user_guide.pdf

regions_www/m/pct_shiny_funs.R
AnnaGoodman1: Ah yes, so you can :) have done that

AnnaGoodman1:
the guidelines are here (esp page 36) https://www.gov.uk/government/uploads/system/uploads/attachment_data/file/472700/NPD_user_guide.pdf , also summarised by me in the attached 
[180207_ApplyingSDCtoschools.docx](https://github.com/npct/pct-shiny/files/1713228/180207_ApplyingSDCtoschools.docx).

I agree about it being better to manipulate first and then we can share the data, which is what I have done in 'school_public' (see e.g. https://github.com/npct/pct-outputs-regional-notR/blob/master/school_public/lsoa/isle-of-wight/z_attributes.csv)

the problem is that I think we do need the *server* ( but not the public) to be able to see the private version of the R files (e.g. https://github.com/npct/pct-outputs-regional-R/blob/master/school/lsoa/isle-of-wight/z.Rds).  That allows it to plot the zones with the correct % cycling colour, but on the pop-up suppress the exact number of children involved.  Otherwise it just has those zones as grey and missing, as per the picture below.  Ditto instead of plotting the route network segment but suppressing the exact number of cyclists, it just omits that segment.  

image


So I think it is about finding some not-too-painful way to get these on the server without having them available to the public if they happen to start exploring our git repos.  Options I can see would include:
1) upload these files direct to the server, bypassing the usual git/lfs method
2) modify the code base in server.R such that instead of expecting to find the 'to_plot' input data in pct-outputs-regional-R/school it instead expect it in a new private repo pct-outputs-regional-R-private/school
3) make the pct-outputs-regional-R/school folders password protected.

3) sounds easiest if feasible in terms of not complicating the codebase/our data handling processes – although maybe something else

regions_www/tabs/region_stats/school/lsoa/avon/region_stats.html
AnnaGoodman1: oh god above - where did he go?! sigh...having a look... :/

regions_www/m/pct_shiny_funs.R
Robinlovelace: Aha, yes I assumed we were using the tidyverse as I use it everyday but agree better to keep it explicit in this case. Thanks for trying and sorry if its been a waste of time (hopefully an interesting waste of time hehe).

regions_www/tabs/region_stats/school/lsoa/avon/region_stats.html
AnnaGoodman1: ok, so, I am pretty sure it is because I a) tweaked the inclusion criteria for the national dataset of children included in the layer and the nb) re-ran the random section of the code in which the 0.3% children with an unknown home LSOA are probabilistically allocated the home LSOA of another child in the same school using the same mode.  I do use seeds in step b), so it does give the same result each time if the input dataset is the same, but did not in this case because the input dataset had been altered by my upstream tweaking in step a). [that sort of modification should not be needed any more once we have agreed the methods for the schools layer, which will be before the final layer goes live]

The upshot was that one unknown-home-LSOA child previously allocated Avon has now been allocated to Somerset – ditto for Gloucestershire. I'm relieved to have found them!

nikolai-b:
@AnnaGoodman1 could we not use the public data on the server and when the server loads the Rds file replace the NA with 5 (or 2 if we are dealing with the route network)?  Would that also work with Dutch or ebike where we the numbers will be > 5 but might allow people to infer backwards?

This would mean we don't have to deal with any potential leaks e.g. the [width](https://github.com/npct/pct-shiny/blob/73771eed21c2b644fefd658cf4cec544b52d9f7b/regions_www/m/server.R#L457) is plotted as continuous variable so could be used to infer more precise count of cyclists.

I would far prefer a way of manipulating the public data but we could quite easily put the files into a password protected `.zip` file, if we choose a long password (~ 30 characters and symbols) then that is secure and unzip the files on the server as part of the update process.

regions_www/tabs/region_stats/school/lsoa/avon/region_stats.html
nikolai-b: That sounds pretty comprehensive search for the lost child.
Thank you for the explanation, it is a shame the 0.3% cause us (mainly you) such grief but it is good that I know some more of the process.

AnnaGoodman1:
Ok thanks @nikolai-b , yes I see what you mean – that does seem good. It seemed easier/more streamlined to me to institute this in pct-scripts rather than as an additional bit of intermediate code between github and the server. this does mean we use up a bit more git LFS storage space, but has the advantage that it gives people access via git to the files that would let them run the schools layer locally.  So now there are three school folders:

- 'school_private' = raw data, local on my machine and not on Github. [I think this is needed as otherwise on the region stats files there will sometimes be small numerical discrepancies that could cause a 'we lost one' (!) headache if anyone queried them - e.g. the number of children the region stats file states live in an area does not perfectly match with the number one gets from the local authority data download]
- 'school' = server data, with z and rnet values of 1 or 2 replaced with 1.5, and d values of 1 to 5 replaced with 3
- 'school_download' (formerly 'school_public') = versions of the datasets that are ready for downloading, with z and rnet values of 1 or 2 replaced with NA, and d values of 1 to 5 replaced with NA.  [NB it seemed to me much easier to save these explicitly rather than try to do it on the fly from the 'school' datasets - not least because identifying which cells need to protected is not completely straightforward for the scenario values]

I have now updated these folders on git, which I think solves this problem.  For now I have only uploaded the Isle of Wight files, I am not keen to upload all files for all regions until DFT formally commissions this work – but they are ready to go on my laptop when that happens.

Issue #661, opened 2018-06-14
Fix IMD tiles


Robinlovelace:
Any ideas @oobrien ? Not an urgent one but thought it worth flagging as I noticed yesterday the tiles are all 1 colour as illustrated in the screenshot below.

Many thanks for providing these in any case, have been very useful to many people.

Robinlovelace:
![image](https://user-images.githubusercontent.com/1825120/41420021-626a6894-6feb-11e8-8c82-b38469048be6.png)

oobrien:
Sorry to have broken this for you. I know it's not the first time I have broken the link for you by moving the tiles. We had an urgent space issue on the server, which is v information used for multiple projects, so we've had to quickly move them. They are now at https://cdrc-maps.liv.ac.uk/tiles/[tileset-name]/...

oobrien:
P.s. I will attach a note in the directory concerned so that you are warned in the future if we have to do so again.

Robinlovelace:
Many thanks for supplying the updated URL (and confirmation that it will not change for the foreseeable offline) @oobrien. And thanks for changing the url @AnnaGoodman1. Confirmation that the layer is now working on production - great stuff!

![image](https://user-images.githubusercontent.com/1825120/50982152-6de7e300-14f4-11e9-86b1-4c9a6649924b.png)

PR #667, opened 2019-01-10
fix IMD url

AnnaGoodman1:
update url, as referenced here https://github.com/npct/pct-shiny/issues/661


AnnaGoodman1:
@Robinlovelace , want to approve this one?

usr110:
Great work, @AnnaGoodman1 

Looks great: 
image

Let me approve it.
Similar change should be done on the _production_ branch as well. I will test it first and then push onto it.

PR #668, opened 2019-01-12
1901 nearmkt - DO NOT MERGE YET!

AnnaGoodman1:
Code for review (don't merge yet!) for new national build and adding near market scenario


AnnaGoodman1:
@usr110 , @Robinlovelace  and @nikolai-b : these are the changes that I think will allow the nearmarket scenario to appear, alongside the latest national build when we launch that.  Great if you want to do any checking, in particular @usr110 I try to modify your LA map but don't totally understand the syntax so you will want to check that.

One difficulty you will have in checking is that you need underlying data files that have the relevant govnearmkt variables. I can't put them on Github as otherwise that would make the new national build/new scenario live I think - but here is a dropbox link to the MSOA files for the Isle of Wight which you could download remotely. Or tell me if there is a better solution... https://www.dropbox.com/sh/12dk0tt3dw0y6o4/AADKcO9kMEt4k_8gVBbfhWmta?dl=0 (these files are in file path pct-outputs-regional-R\commute\msoa\isle-of-wight)

Robinlovelace:
Great work Anna. We can test locally no problem. Great to see the near market scenario 'near market'.

usr110:
Great and thanks a bunch, @AnnaGoodman1 !
Looking at it. Will later on have a look at the `la-map`

usr110:
@AnnaGoodman1 it seems to work fine for `isle-of-wight` - there is no technical problem loading or switching between scenarios, so well done!

Re looking at the specific numbers, I see that at the macro level, there's not much difference between `Government Target (equity)` and `Government Target (near market)`, as you'd expect. However when you look at specific desire-lines, you'd see that with the same increase of cyclists in a particular flow, the values for `Change in deaths/yr` and `Change in CO2e (t/yr)` are slightly different. It may be a rounding issue. See below an example:


Scenario: Government Target (near market)
Isle of Wight 011 - Isle of Wight 017
------------ | -------------
Total commuters: | 220
Cyclists (baseline): | 6 (3%)
Cyclists (scenario): | 11 (5%)
Change in drivers: | -4
Change in deaths/yr: | **-0.002 (£4499)**
Change in CO2e (t/yr): | **-1.2**
Distance (km): | 9

Scenario: Government Target (equity)
Isle of Wight 011 - Isle of Wight 017
------------ | -------------
Total commuters: | 220
Cyclists (baseline): | 6 (3%)
Cyclists (scenario): | 11 (5%)
Change in drivers: | -4
Change in deaths/yr: | **-0.003 (£5847)**
Change in CO2e (t/yr): | **-1.3**
Distance (km): | 9

AnnaGoodman1:
Thanks @usr110 ! And yes, that is right - it is not a rounding issue, it is to do with the fact that in the near market scenario you are disproportionately switching young people, since those are the ones who cycle most. Since young people are also more likely previously to have been walking, the near market scenario tends to generate fewer physical activity and carbon benefits. The fact that they are younger, and therefore lower baseline risk, also contributes further to the health benefits being smaller.

usr110:
I see, thanks for the explanation, @AnnaGoodman1

Robinlovelace:
Looks good to me @AnnaGoodman1 I can test and merge Tuesday to see this on the test server, hopefully with the new tiles built by @nikolai-b ready by then. Sound like a plan? Thanks!

nikolai-b:
@AnnaGoodman1 I've made a few changes.  The front page uses `pct_regions_lowres_scenario_commute.js` (not the geojson).  I've no idea how that is generated but it looks like you managed so I updated the `.js` file and deleted the confusing `geojson` ones.  I've also reverted the changes to the LA map, that uses different data which hasn't had the near market scenario added (again, I'm not sure how the data was generated).  Also that map isn't used on our site yet and it is used on https://www.theguardian.com/environment/bike-blog/2017/apr/13/how-much-could-commuter-cycling-increase-in-your-part-of-england.  If we do intend to add a link to the la-map (as per #671) then I'd like it to share as much code and data from the front page map as possible so we can fix it then if needs be.  Please let me know if you disagree though, I just noticed it wasn't working.

I've also updated the code based from the master branch, hopefully that is ok.

PR #693, opened 2019-03-08
Front page updates

nikolai-b:
@usr110 this would makes your new map.  If your happy with the code the it can be merged into your branch.

If you want to check it locally you can replace `regions_www/index.html` with: 
[index.html](https://github.com/npct/pct-shiny/files/2947658/index.html.txt) 
(I've just change resources to be `http` instead of files.)
Note you'll have to rename it as :octocat: doesn't let me upload `.html` files.


usr110:
Great improvements there, and thanks for fixing the bugs. Just tested it and it looks fabulous!
Also the code reads perfectly well. Good job!. 
(in hindsight, should've left this for you)

Just have two minor suggestions:

1) Following the current version of the map, I think the order of the controls was better where we have info box first, and then the scenario selection. Similarly in this new map, I'd propose putting the text box first (where `Hover over a region` appears), and then layer selection (commute or school) and then scenarios.

2) I'd propose changing the title (`cycle to work` or `cycle to school`) of the `hover over a region` box as soon as a user selects a layer. At the moment, it only updates when a user hovers over a region, which I think is not ideal. 

What do you think?

nikolai-b:
@usr110 Yes to both of these.  With 1) then I think info at the top makes a bit more sense so I've moved it back. With 2) I meant to do that but didn't write it down!  Done in https://github.com/npct/pct-shiny/pull/693/commits/6865e0473a8f8e5a11d511ebe9dd7a9c4d8791e3.

> (in hindsight, should've left this for you)

Always good to have multiple :eyes: on it and I find it easier to think things though once there is some working code so I'm very glad you did it!

usr110:
> @usr110 Yes to both of these. With 1) then I think info at the top makes a bit more sense so I've moved it back. With 2) I meant to do that but didn't write it down! Done in [6865e04](https://github.com/npct/pct-shiny/commit/6865e0473a8f8e5a11d511ebe9dd7a9c4d8791e3).

@nikolai-b glad you agreed :). Thanks for implementing them. Just tested it, and it looks fabulous!

> Always good to have multiple eyes on it and I find it easier to think things though once there is some working code so I'm very glad you did it!

Always humble! :+1: 

Will now merge your branch. Thanks again.

Issue #705, opened 2019-04-04
Manchester all trips layer


JDWoodcock:
it shouldnt be showing but it is on pct.bike

Robinlovelace:
@usr110 are you up for taking a look? As I recall it was your great work getting it up. Thanks for spotting that James!

usr110:
Good catch indeed, @JDWoodcock 

Have now created a PR for it: https://github.com/npct/pct-shiny/pull/706

@Robinlovelace would you mind testing it?

JDWoodcock:
it was good catch by @RachelAldred during training (I think training is really useful for bug spotting!) Im just the messanger

nikolai-b:
This is deployed and all trips is no longer available.  Thanks!

Robinlovelace:
Great work @nikolai-b

PR #706, opened 2019-04-05
Make all trips option available (whereever the data exists) for non-prod branches

usr110:
This is in reference to the issue created here: https://github.com/npct/pct-shiny/issues/705

@Robinlovelace, are you up to test it first, and then merge?

I think it would be wise to merge to `master` and then `prod` afterwards. What do you think?

### DIFF
--- a/regions_www/m/server.R
+++ b/regions_www/m/server.R
@@ -291,6 +291,8 @@ shinyServer(function(input, output, session) {
     # Identify purposes and geographies available in region
     purposes_list <- c("commute", "school", "alltrips")
     new_purpose <- (dir.exists(file.path(data_regional_root, purposes_list, "msoa", region$current)) | dir.exists(file.path(data_regional_root, purposes_list, "lsoa", region$current)))
+    # Remove alltrips from new_purpose for the production branch, even if the data directory exists
+    new_purpose <- ifelse(is_prod, new_purpose[new_purpose != "alltrips"], new_purpose)
     if(!identical(new_purpose,region$purposes_present)){
       region$purposes_present <- new_purpose
     }
--- a/regions_www/m/server.R
+++ b/regions_www/m/server.R
@@ -41,8 +41,8 @@ if (!on_server) {
 
 repo_sha <-  as.character(readLines(file.path(interface_root, "repo_sha")))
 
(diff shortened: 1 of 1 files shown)


Robinlovelace:
Thanks Ali, will test when I get a chance on work desktop - likely next week now.

Robinlovelace:
Update: just looked at the code, looks good to me, will test on a local copy of prod. @nikolai-b can you see any issues with this? Good fix by the looks of it @usr110 many thanks.

regions_www/m/server.R
```diff
@@ -41,6 +41,9 @@ if (!on_server) {
 
 repo_sha <-  as.character(readLines(file.path(interface_root, "repo_sha")))
 
+# Check if the branch is production or not
+is_prod <- ifelse(system("git symbolic-ref -q --short HEAD", intern = T) == "production", TRUE, FALSE)
```
nikolai-b: The code isn't inside a git repo in production so sadly this doesn't work:

```bash
root@npct0:/var/shiny/pct-shiny# git symbolic-ref -q --short HEAD
fatal: Not a git repository (or any of the parent directories): .git
```
Which I think you spotted first https://github.com/npct/pct-shiny/commit/76cd62e40e103e29f64d946f6f96bb10aa5ea710

I believe we are after
```R
# Check if we are on the production server (npt followed by any number of digits (only) is a prod machine)
production_branch <- grepl("npt\\d*$", Sys.info()["nodename"])
```
(from deb05c6)

usr110: Very good point @nikolai-b  - thanks!
Will make the required changes.  So glad you checked it before it got merged.
I had a recollection that we did something like this before, but couldn't find the commit - may need to learn more about git, it seems.

nikolai-b: Sure!

I used [git log -S](https://git-scm.com/docs/git-log#Documentation/git-log.txt--Sltstringgt) with `git log -S production` to list relevant commits then I just read the commit messages and used `git show`

usr110: Noted, thanks!

usr110:
Thanks @nikolai-b  for the merge!

Issue #707, opened 2019-04-05
Reduce memory usage


nikolai-b:
Each time a new Rds file is loaded it takes memory, if 10 people look at the same region we open 10 copies of that Rds file.

https://shiny.rstudio.com/articles/scoping.html

It is possible to share this memory but loaded all the Rds files isn't possible (takes ~ 14 Gb and our server only has 8 Gb)

Robinlovelace:
This has been fixed so closing. One danger I can see with this is that if 'bad' data gets loaded, it will persist in the global environment for a long time. I think this is currently happening for http://npct0.vs.mythic-beasts.com/m/?r=cheshire. Thoughts @nikolai-b ? Amazing memory-reducing work in any case, and should not have any negative consequences once all the data is 'good'.

PR #708, opened 2019-04-05
Memory

nikolai-b:
Ref #707 

This isn't ready but it shows the basic idea.

### DIFF
--- a/regions_www/m/server.R
+++ b/regions_www/m/server.R
@@ -228,14 +228,31 @@ shinyServer(function(input, output, session) {
   helper$old_geog <- ""
   helper$old_purpose <- ""
 
-  load_data <- function(filepath){
+  load_data <- function(base_path, filename, purpose, str_lines = NULL){
+    filepath <- file.path(base_path, filename)
+
     while (length(loaded_data) > 100) { # Rm objects from the list when too many (by time last accessed)
       loaded_data[[names(loaded_data_accessed[loaded_data_accessed == min(unlist(loaded_data_accessed))])]] <<- NULL
     }
     if (file.exists(filepath)) {
       loaded_data_accessed[[filepath]] <<- Sys.time()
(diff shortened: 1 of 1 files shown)


nikolai-b:
I navigated to Hampshire on multiple different browser windows and recorded memory usage in R by `pryr::mem_used()`

Old code | Sessions | MB | Newer code (before 78db7c8) | Sessions | MB | Newest code | Sessions | MB
-- | -- | -- | -- | -- | -- | -- | -- | --
  | 1 | 166 |   | 1 | 168 |   | 1 | 185
  | 2 | 272 |   | 2 | 213 |   | 2 | 185
  | 3 | 378 |   | 3 | 259 |   | 3 | 186
  | 4 | 484 |   | 4 | 304 |   | 4 | 186
  | 5 | 590 |   | 5 | 351 |   | 5 | 187
  | 6 | 696 |   | 6 | 396 |   | 6 | 188

So now you can see multiple sessions adds hardly anything to the total MB of R objects!

I've added a cap so that objects are removed from the global list if there are more than 100 and they are removed last accessed == first removed.  This is because I don't want that global list to get too large (loading all the rds files is ~ 15Gb).


This will need testing as I've also made `to_plot` a `reactiveValue` instead of it being a global object so lots of the observes have changed.

cc @Robinlovelace @usr110

Robinlovelace:
Thanks @nikolai-b I just took a look at the code, looks clever. Look forward to testing this when back home next week. Have asked a few questions in the meantime. I suggest that if it passes informal tests by me and @usr110 we get others to test it on the test server with a before/after merge. Sound reasonable? I think this could make the app way more tolerant of many users so great work!

Robinlovelace:
I've tested on the latest version with the new tiles in all scenarios for commuting and schools and for multiple regions. Aside for making large regions load faster, I think, I can see no change. So merging. Great work @nikolai-b, many thanks!

PR #713, opened 2019-04-17
Near market map

nikolai-b:
:wave: you code works fine @usr110 but I think we can do this a bit cleaner, WDYT?

Also can we go back to making Pull requests please?  I know I've recently pushed to master but I think it is much better if we always get a review.


Robinlovelace:
Agreed. We should use PRs.

usr110:
> wave you code works fine @usr110 but I think we can do this a bit cleaner, WDYT?

Great work there @nikolai-b :+1: 


> Also can we go back to making Pull requests please? I know I've recently pushed to master but I think it is much better if we always get a review.

Fair point. Will use PR now.

PR #720, opened 2019-04-19
Fix grid on front page

### LINKED ISSUE #702: Map on landing page has square outlines on firefox
See below. Can anyone else reproduce this behaviour? I noticed it after recent updates to production

image

Looking at the latest updates in https://github.com/npct/pct-shiny/commit/820663cd8eb47240f84ded83e768482767d5b6f2 I cannot see anything obvious though. Ideas @nikolai-b and @usr110 ?

nikolai-b:
This appears to be a bug in leaflet so bumping the version

Fixes #702

I can see that this was fixes in version 1.0.1 so likely it is caused by https://github.com/Leaflet/Leaflet/pull/4956 but :man_shrugging: .  Checking the leaflet changelog it seems safe to upgrade.


usr110:
Brilliant fix @nikolai-b :+1:

PR #723, opened 2019-04-29
Rename equality

AnnaGoodman1:
rename equity to equality


AnnaGoodman1:
can you do this @Robinlovelace or @nikolai-b  ? and then @nikolai-b merge to productin

PR #730, opened 2019-05-01
Fix disabling controls

### LINKED ISSUE #729: Landing page map: layers are not turn off based on the selected scenario
So glad that the [radio button bug](https://github.com/npct/pct-shiny/issues/726) was resolved - good job there @nikolai-b. Just testing it a bit more produced another bug. The travel layer (either `commute` or `school`) is not reactive based on the scenario a user has selected. So for instance it should disable `school` layer for `Gender Equality` scenario, which it doesn't. To reproduce it, select `commute` layer, and then select `Gender Equality` - this should disable `school` layer, but it doesn't. See below:

image

nikolai-b:
Use the https://api.jquery.com/attribute-starts-with-selector/ so we select the correct radio buttons

Fixes #729

### DIFF
--- a/regions_www/www/front_page/map.js
+++ b/regions_www/www/front_page/map.js
@@ -63,7 +63,7 @@ $(document).ready(function(){
 
   L.control.layers(commuteLayers, null, {collapsed: false}, {position: 'topleft'}).addTo(map);
   defaultCommuteLayer.addTo(map);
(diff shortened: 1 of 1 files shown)


Robinlovelace:
Great fix, ready to go onto production.

Before:

image

After:

image

PR #747, opened 2019-06-27
Limit loaded regions by size

### LINKED ISSUE #746: Server ran out of memory
See #742 for background but makes sense to have the memory leak as a separate issue.

I added a global array which stores regions but once 100 regions have been loaded they [should be removed](https://github.com/npct/pct-shiny/blob/2714d4dc758d13987891d1622d24a67bc9b1d477/regions_www/m/server.R#L243-L245).  Maybe that doesn't free up memory @Robinlovelace ?  Maybe something else isn't working as expected...

nikolai-b:
Fixes #746

### DIFF
--- a/regions_www/m/server.R
+++ b/regions_www/m/server.R
@@ -240,7 +240,7 @@ shinyServer(function(input, output, session) {
   load_data <- function(base_path, filename, purpose, str_lines = NULL){
     filepath <- file.path(base_path, filename)
     while (format(object.size(loaded_data), units = "Gb") > 3 ) { # Rm objects (by time last accessed) if the list size is more than 3 Gb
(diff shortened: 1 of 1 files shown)


nikolai-b:
@Robinlovelace thanks, 

1. We have two R instances running on each server, so 3 Gb x 2 = 6 Gb and I wanted to leave space for the other things...
1. Quite possibly, I felt that line was a bit clumsy, I'll check tomorrow :golf: 
1. No idea about R's garbage collector, I naively assumed that releasing the memory to the OS is pointless as this function is: 1st checking if the array is too big (and removes something from it if it is), 2nd it loads a new Rds file.  So any memory returned to the OS in 1st step will then be consumed by R in the 2nd.  I've not done any digging so I could easily be very very wrong here :smile: ...

Robinlovelace:
Tested locally and works. Obviously hard to test this in a production setting with multiple users accessing GB of data, but confident this is an improvement so merging.

PR #751, opened 2019-07-03
Deployment update

### LINKED ISSUE #742: App ran out of memory and it did not restart
When @RachelAldred was demonstrating the pct it ran out of memory (RAM) it exited with an error but it was not automatically restarted #740 

### ~~Why did it run out of memory~~

See #746

### Why did it not restart?

Systemd monitors shiny-server and should restart it if it fails.  We are using a modified version of [shiny-server.service](https://github.com/rstudio/shiny-server/blob/v1.5.9.923/config/systemd/shiny-server.service) that is old.  We should update it and it should live inside this repo.

nikolai-b:
Fixes #742 

This adds most of the scripts that are on the prod server but moves them inside this git repo to allow us to change them on deploy.  I've tested this marginally but will need to monitor the next deployments carefully.

### DIFF
--- a/Server.md
+++ b/Server.md
@@ -7,7 +7,7 @@ to draw pretty pictures of how often people cycle.
 
 You can see the data for Cambridge here,
 
-http://pct.bike/cambridgeshire/
+http://www.pct.bike/cambridgeshire/
 
 ## Servers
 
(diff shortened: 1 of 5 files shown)


Server.md
```diff
@@ -7,7 +7,7 @@ to draw pretty pictures of how often people cycle.
 
 You can see the data for Cambridge here,
 
-http://pct.bike/cambridgeshire/
+http://www.pct.bike/cambridgeshire/
```
Robinlovelace: Minor point, but wouldn't that be better as:

```
https://www.pct.bike/cambridgeshire/
```

PR #756, opened 2019-07-10
Deploy

nikolai-b:
This deployment has:

- The rq fixed data
- Fix to hide the "Order lines by" select when there are no lines
- Better memory handling
- Large deployment script update (including better restart handling of when Shiny server crashes)


nikolai-b:
As this has new deployment scripts I'd like to be around when it deploys so if some is happy and approves would you mind leaving me to merge?  Thanks!

PR #757, opened 2019-07-11
Production

nikolai-b:
Merge prod back into master.  I'll try amend the settings so this isn't needed


Robinlovelace:
Can we close this @nikolai-b  ? Seems no files have changed.

nikolai-b:
I'd like them to be in sync. I know no files have changed but after merging into production it is good to merge back into master. That way they both have the same SHAs

PR #758, opened 2019-07-12
Fix issue with deployment script

nikolai-b:
I noticed two issues when we deployed, this fixes them:

Fix invalid perl, tweak systemd as we were getting:
Jul 12 18:48:43 npt2 systemd[1]: shiny-server0.service: Supervising process 465 which is not our child. We'll most likely not notice when it exits.


Robinlovelace:
Looks good, but suggest we merge Wednesday, after the training course.

nikolai-b:
I believe this caused https://github.com/npct/pct-shiny/issues/759 

Like @usr110 suggested I think this happened because one of the servers was in an inconsistent state due to this bug.  I've fixed it on the server and will re-deploy after Wed

PR #761, opened 2019-07-17
[HOTFIX] Fix removing loaded objects

nikolai-b:
Need to use single [ or we get `attempt to select less than one element in integerOneIndex`
Need to also remove the element from loaded_data_accessed

### DIFF
--- a/regions_www/m/server.R
+++ b/regions_www/m/server.R
@@ -239,8 +239,10 @@ shinyServer(function(input, output, session) {
 
   load_data <- function(base_path, filename, purpose, str_lines = NULL){
     filepath <- file.path(base_path, filename)
-    while (format(object.size(loaded_data), units = "Gb") > 3 ) { # Rm objects (by time last accessed) if the list size is more than 3 Gb
-      loaded_data[[loaded_data_accessed == min(unlist(loaded_data_accessed))]] <<- NULL
+    while (format(object.size(loaded_data), units = "Gb") > 2) { # Rm objects (by time last accessed) if the list size is more than 2 Gb
+      idx_to_remove <- loaded_data_accessed == min(unlist(loaded_data_accessed))
+      loaded_data[idx_to_remove] <<- NULL
+      loaded_data_accessed[idx_to_remove] <<- NULL
     }
     if (file.exists(filepath)) {
       loaded_data_accessed[[filepath]] <<- Sys.time()


nikolai-b:
Note this is merging straight into production as it is down anyway :cry:

regions_www/m/server.R
```diff
@@ -239,8 +239,10 @@ shinyServer(function(input, output, session) {
 
   load_data <- function(base_path, filename, purpose, str_lines = NULL){
     filepath <- file.path(base_path, filename)
-    while (format(object.size(loaded_data), units = "Gb") > 3 ) { # Rm objects (by time last accessed) if the list size is more than 3 Gb
-      loaded_data[[loaded_data_accessed == min(unlist(loaded_data_accessed))]] <<- NULL
+    while (format(object.size(loaded_data), units = "Gb") > 2) { # Rm objects (by time last accessed) if the list size is more than 2 Gb
+      idx_to_remove <- loaded_data_accessed == min(unlist(loaded_data_accessed))
+      loaded_data[idx_to_remove] <<- NULL
```
Robinlovelace: Do you not want double bracket notation here?

nikolai-b: Did you read the MR description?  i.e.

```r
loaded_data <- list()
loaded_data_accessed <- list()
filepath = "a"
loaded_data_accessed[[filepath]] <- Sys.time()
loaded_data[[filepath]] <- "big file a"
filepath = "b"
loaded_data_accessed[[filepath]] <- Sys.time()
loaded_data[[filepath]] <- "big file b"
loaded_data[[loaded_data_accessed == min(unlist(loaded_data_accessed))]] <- NULL
```

Gives 
`Error in loaded_data[[loaded_data_accessed == min(unlist(loaded_data_accessed))]] <- NULL : 
  more elements supplied than there are to replace`

Robinlovelace: Also, I think this would be better:

```
loaded_data[idx_to_remove] <- NULL
```

Because that object is already present in the environment.

nikolai-b: I used single bracket because you suggested it https://github.com/npct/pct-shiny/pull/747 and I thought I tested it but now it looks like it doesn't work, no idea why though, you can probably tell me :smile:

nikolai-b: I used global assignment as I want to make sure it is removed from the global list.  If it is only removed from the present environment then it will still persist in the global and will still be using memory

Robinlovelace: Reprex to illustrate the point (`x <<- NULL` seems to fail, although `[` and `[[` both seem to work, the latter is better I think):

``` r
l = list(a = 1, b = 2)
l
#> $a
#> [1] 1
#> 
#> $b
#> [1] 2
l[1] 
#> $a
#> [1] 1
l[[1]]
#> [1] 1

l[[1]] <- NULL
l
#> $b
#> [1] 2


l = list(a = 1, b = 2)
l[1] <- NULL
l
#> $b
#> [1] 2

l = list(a = 1, b = 2)
l[1] <<- NULL
#> Error in l[1] <<- NULL: object 'l' not found
l
#> $a
#> [1] 1
#> 
#> $b
#> [1] 2
```

<sup>Created on 2019-07-17 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Robinlovelace: One to discuss later. I'm sure reprexes like mine above will help.

nikolai-b: @Robinlovelace is there a reason you didn't start with https://github.com/npct/pct-shiny/pull/761#discussion_r304509409 as follows the actual code more?

In your example you are passing in an integer to [], i.e. `l[1]` whereas in the code we pass in a list `loaded_data_accessed == min(unlist(loaded_data_accessed))`, could that be why we need single []?

As for global assignment it is needed because we are inside a function and want to change the global variable, here is a more full example:

```r
loaded_data <<- list()
loaded_data_accessed <<- list()
load_data <- function(fp){
  loaded_data_accessed[[fp]] <<- Sys.time()
  loaded_data[[fp]] <<- paste("big file", fp)
}
remove_data <- function(){
  loaded_data[loaded_data_accessed == min(unlist(loaded_data_accessed))] <<- NULL
  loaded_data_accessed[loaded_data_accessed == min(unlist(loaded_data_accessed))] <<- NULL
}
load_data("a")
load_data("b")
length(loaded_data)
remove_data()
```

again this gives me errors with single brackets [, and it doesn't remove or add things properly without `<<-`

Robinlovelace: There was, didn't realise that was a reprex. Works good for illustrative purposes:

``` r
loaded_data <<- list()
loaded_data_accessed <<- list()
load_data <- function(fp){
    loaded_data_accessed[[fp]] <<- Sys.time()
    loaded_data[[fp]] <<- paste("big file", fp)
}
remove_data <- function(){
    loaded_data[loaded_data_accessed == min(unlist(loaded_data_accessed))] <<- NULL
    loaded_data_accessed[loaded_data_accessed == min(unlist(loaded_data_accessed))] <<- NULL
}
load_data("a")
load_data("b")
loaded_data
#> $a
#> [1] "big file a"
#> 
#> $b
#> [1] "big file b"
length(loaded_data)
#> [1] 2
remove_data()
length(loaded_data)
#> [1] 1
```

<sup>Created on 2019-07-17 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Robinlovelace: And it's true, the `<<-` operator is needed when in a function: 

``` r
loaded_data <<- list()
loaded_data_accessed <<- list()
load_data <- function(fp){
  loaded_data_accessed[[fp]] <<- Sys.time()
  loaded_data[[fp]] <<- paste("big file", fp)
}
remove_data <- function(){
  loaded_data[loaded_data_accessed == min(unlist(loaded_data_accessed))] <- NULL
  loaded_data_accessed[loaded_data_accessed == min(unlist(loaded_data_accessed))] <- NULL
}
load_data("a")
load_data("b")
loaded_data
#> $a
#> [1] "big file a"
#> 
#> $b
#> [1] "big file b"
length(loaded_data)
#> [1] 2
remove_data()
length(loaded_data)
#> [1] 2
```

<sup>Created on 2019-07-17 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Robinlovelace: Tested and also load times with this code:

```
    # while (format(object.size(loaded_data), units = "Gb") > 2) { # Rm objects (by time last accessed) if the list size is more than 2 Gb
    #   message("High mem usage, deleting")
    #   idx_to_remove <- loaded_data_accessed == min(unlist(loaded_data_accessed))
    #   loaded_data[idx_to_remove] <<- NULL
    #   loaded_data_accessed[idx_to_remove] <<- NULL
    # }
    # message("Mem use after:", format(object.size(loaded_data), units = "Mb"))
    if (file.exists(filepath)) {
      loaded_data_accessed[[filepath]] <<- Sys.time()
      if (is.null(loaded_data[[filepath]])) {
      t1 = Sys.time()
        rds <- readRDS(filepath)
      t2 = Sys.time()
```

18 s to load London when not pre-loaded, around 9 s after.

Issue #759, opened 2019-07-15
Server reliability


JDWoodcock:
hi all
not sure best way to notify but given potential issues to send out reminder there is pct training tomorrow so hopefully will run smoothly
@nikolai-b

JDWoodcock:
on that I am noticing some problems loading for London as is Ali
The application unexpectedly exited.


Diagnostic information is private. Please ask your system admin for permission if you need to check the R logs. shiny-server-client.min.js:1:24931
Mon Jul 15 2019 14:39:28 GMT+0100 (British Summer Time) [INF]: Connection closed. Info: {"type":"close","code":4503,"reason":"The application unexpectedly exited","wasClean":true} shiny-server-client.min.js:1:21781
Mon Jul 15 2019 14:39:28 GMT+0100 (British Summer Time) [DBG]: SockJS connection closed shiny-server-client.min.js:1:5353
Mon Jul 15 2019 14:39:28 GMT+0100 (British Summer Time) [DBG]: Channel 0 is closed shiny-server-client.min.js:1:5353
Mon Jul 15 2019 14:39:28 GMT+0100 (British Summer Time) [DBG]: Removed channel 0, 0 left shiny-server-client.min.js:1:5353

JDWoodcock:
ok now getting this for Cambridgeshire
Mon Jul 15 2019 15:52:16 GMT+0100 (British Summer Time) [INF]: Connection opened. http://pct.bike/m/?r=cambridgeshire shiny-server-client.min.js:1:21781
Mon Jul 15 2019 15:52:16 GMT+0100 (British Summer Time) [DBG]: Open channel 0 shiny-server-client.min.js:1:5353
The application unexpectedly exited.



Diagnostic information is private. Please ask your system admin for permission if you need to check the R logs. shiny-server-client.min.js:1:24931
Mon Jul 15 2019 15:52:17 GMT+0100 (British Summer Time) [INF]: Connection closed. Info: {"type":"close","code":4503,"reason":"The application unexpectedly exited","wasClean":true} shiny-server-client.min.js:1:21781
Mon Jul 15 2019 15:52:17 GMT+0100 (British Summer Time) [DBG]: SockJS connection closed shiny-server-client.min.js:1:5353
Mon Jul 15 2019 15:52:17 GMT+0100 (British Summer Time) [DBG]: Channel 0 is closed shiny-server-client.min.js:1:5353
[...]

PR #763, opened 2019-07-18
Round null

nikolai-b:
Should fix https://github.com/npct/pct-shiny/issues/759#issuecomment-512853445

### DIFF
--- a/regions_www/m/pct_shiny_funs.R
+++ b/regions_www/m/pct_shiny_funs.R
@@ -40,6 +40,9 @@ normalise <- function(values, min = 0, max = 1){
 
 ## Create Values such that rounded to XDP if >0 and <0.5.  [NB can be NaN if denominator zero]
 round_dp <- function(expression, small_threshold = 0.05, large_threshold = 0.5){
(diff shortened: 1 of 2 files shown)


nikolai-b:
Agreed but this is bringing the site down.  We can make this change, fix the data, then revert this.  I very much want to fix whatever this issue is upstream but once it brings down production it seems more important that we fix the site.

Robinlovelace:
If you're sure that's the cause. I only checked 1 log file. Have you checked more? Approving in any case as cannot see any downsides. Thanks for clarification.

nikolai-b:
@Robinlovelace found out what this is, when on Cambridge scenario and we change back to commute then the input shiny state can be "cambridge" and "commute" briefly before the "cambridge" scenario is deselected, hence the crash.  I think it is fine to then allow this to ignore NULL otherwise we would have to start messing with observer priority https://shiny.rstudio.com/reference/shiny/latest/observe.html and I'm not keen on that.

Robinlovelace:
Agreed. Good diagnosis.

PR #769, opened 2019-07-28
Newmanual

AnnaGoodman1:
new manual


AnnaGoodman1:
why don't we use this as a place to put any last changes, @Robinlovelace - save endless pull requests...

Robinlovelace:
Sounds good @AnnaGoodman1 we can make any last touches here. Will hold fire on merging, but the changes look fine, so approving.

PR #770, opened 2019-07-29
Fix rounding issue on staging

nikolai-b:
This was merged into prod but not yet into staging @Robinlovelace

### DIFF
--- a/regions_www/m/server.R
+++ b/regions_www/m/server.R
@@ -30,7 +30,7 @@ must_be_installed_pkgs <- c("rgdal", "rgeos", "shinyjs")
 interface_root <- file.path("..", "..")
 data_regional_root <-  file.path(interface_root, '..', 'pct-outputs-regional-R')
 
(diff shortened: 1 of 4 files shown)


regions_www/m/server.R
```diff
@@ -30,7 +30,7 @@ must_be_installed_pkgs <- c("rgdal", "rgeos", "shinyjs")
 interface_root <- file.path("..", "..")
 data_regional_root <-  file.path(interface_root, '..', 'pct-outputs-regional-R')
 
-outputs_regional_sha <- as.character(readLines(file.path(interface_root, "outputs_regional_sha")))
+outputs_regional_sha <- as.character(readLines(file.path(interface_root, "outputs_regional_sha"), warn = F))
```
Robinlovelace: Minor comment, but it's seen as best practice to use `FALSE` instead of `F`.

Robinlovelace:
Good to have the two versions as close as possible to each other, forgot we pushed directly to production last time.

PR #774, opened 2019-07-30
Cambridge crash

nikolai-b:
I've commented in the code but this is easy to reproduce the crash:

* select Schools, Go Cambridge and Route Network (Clickable)
* select Commute
* Shiny dies :skull:


usr110:
I think we should give a bit more time for testing before we merge PRs like this @Robinlovelace and @nikolai-b 


(case in point: https://github.com/npct/pct-shiny/issues/775)

Robinlovelace:
This fixed a serious bug (crashes) so I think it was the right decision for @nikolai-b, even if we knew it would have this unintended consequence. But totally agree, time for testing is a good thing, noted.

PR #776, opened 2019-07-31
Use slc instead of sivalue_heat for route_network. Fix #775

usr110:
As the description says. Relevant issue: #775.

PR #777, opened 2019-07-31
Always use slc to test if the data is missing

nikolai-b:
Follow up to #776 

@usr110 very good point.  I hadn't checked if all the types of data have `sivalue_heat`.  I checked and they all do have `slc` so we can always use that.


usr110:
Great @nikolai-b. Seems like a good solution. Let me approve it.

Issue #781, opened 2019-08-02
Speed-up load times


Robinlovelace:

AnnaGoodman1:
If the input datasets were ,say,50% smaller then would we expect website loading speeds to be roughly x2 faster?  Should we experiment? One could reduce the input dataset size by e.g.

1.	Reducing the maximum length of lines – particularly since longer lines take more memory. Going from a maximum line length of 20km to 10km gets the following improvements in .Rds file size:
•	avon: 7.3 to 2.3MB
•	london 79.6MB to 37.3MB

2.	Reducing the number of attribute variables saved, e.g. I think there are 54/87 variables from the rf.Rds and l.Rds files that are in the downloads because potentially useful but not actually needed for the interface.  But this has much more modest effects
•	avon: 7.3 to 6.73MB
•	london 79.6MB to 73.2MB

3.	Possibly there would also be some other options e.g. around simplifying the rf/rq line shapes. 

In any of these cases, I would propose that the downloads were not changed and these simplifications only affected what one got in the online interface. This could e.g. be done by duplicating the regional R file repo such that the data download page pointed to a 'full version' repo (same as current) and the online interface/server only had a new 'smaller subset' repo.  I appreciate this duplication is not DRY ideal, but is the easiest way I can see to reduce the input dataset size while retaining the quality of the data downloads.

AnnaGoodman1:
Hi all,

So I started having a go at my suggestion, by 
 - creating smaller l/rf/rq files (maximum fast route distance 10 km not 20 km); this roughly halves these file sizes
 - saving these at pct-outputs-regional-Rsmall (https://github.com/npct/pct-outputs-regional-Rsmall)
 - comparing load times on my laptop for the current system versus one that instead direct to these Rsmall files (via branch https://github.com/npct/pct-shiny/tree/test-small-datasets).

The regions that exist in this small state are LSOA versions of Avon, Hampshire, Isle of Wight and London

On my laptop thing seem to get a lot faster for London:

- loading region to get zones showing on screen: from 53 sec to 17 sec
- then loading straight lines: from 16 sec to 8 sec
- then switching to loading rf / rq lines: from 27 sec to 17 sec

@usr110 , @nikolai-b , @Robinlovelace : Do you think this smaller file size approach would plausibly solve our problems?  if so shall we try it more fully, by me generating small files for all regions and then we change what files are synced up on the server by changing the syncing repo?

Robinlovelace:
[...]

PR #788, opened 2019-08-29
Lazy load

nikolai-b:
Re #781

- Speed up freeing memory: Only ever remove one data file at a time 
- Load the straight lines and routes only if needed

This is much faster for me :smile:

### DIFF
--- a/regions_www/m/server.R
+++ b/regions_www/m/server.R
@@ -550,94 +550,100 @@ shinyServer(function(input, output, session) {
 
   }
 
-  ## Plot if lines change
   observe({
-    if (interactive()){
-      start_time <- Sys.time()
-      loading_finish_time <- start_time
-    }
-    shinyjs::showElement(id = "loading")
-
-    line_type <- ifelse(input$line_type == 'routes', "routes_quieter", input$line_type)
-
     if (region$line_data_dir != region$data_dir && (input$line_type %in% c("straight_lines", "routes_fast", "routes")) ){
       region$line_data_dir <- region$data_dir
       region$plot$straight_lines <- load_data(region$data_dir, "l.Rds", input_purpose())
       region$plot$routes_fast <- load_data(region$data_dir, "rf.Rds", input_purpose())
       region$plot$routes_quieter <- load_data(region$data_dir, "rq.Rds", input_purpose(), region$plot$straight_lines)
+      return()
       if (interactive()){
         loading_finish_time <- Sys.time()
       }
     }
+  }, priority = 10)
 
-    local_lines <- sort_lines(region$plot[[line_type]], input$line_type, input$nos_lines)
-
-    # Filter out zero lines for scenario in question from route network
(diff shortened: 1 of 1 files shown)


Robinlovelace:
Fantastic work. No time to test this week, should get on it Tuesday. Could you provide a rough estimate of how much faster? And what kind of usage cycle to you use, many regions I assume. By the sounds of it this + #789 should allow us to fix #781.

nikolai-b:
Speed up depends on a few too many things. If you are changing region and are viewing lines and the `loaded_data` isn't full then this provides no speeds up. If you have no lines or route network selected and `loaded_data` is full and R gets confused and the `while` loop removes everything in `loaded_data` then it could be a +15s speed up.

nikolai-b:
I've added loading and ordering time logging.  In doing that I notice we are doing a the most expensive operations twice (loading a region and sorting) so I've also fixed that.

From https://github.com/npct/pct-shiny/pull/789#issuecomment-527648285 

So currently London MSOA commute with lines or routes takes (on my laptop) ~ 8.3 s to load the Rds files and an additional ~ 2.3 s to order and plot the lines.

With this PR if no lines are selected, or route network then we don't load the route Rds files so the London load time is reduced to ~ 0.2 s (pretty fast).  If we are loading the lines / routes Rds files then with this PR we now save ~ 4 s by only loading the region once and sorting the lines once.  I've tested locally but really make sense for this to be tested again as it changes a few things.

Robinlovelace:
Excellent and important work by the sounds of it @nikolai-b - will test now with an aim to get this on production ASAP.

Robinlovelace:
For testing, it's ~9 seconds on my computer for London. Before this MR:

Peek 2019-09-04 10-41

Robinlovelace:
After the fix it is much faster, ~3s I counted, output from new code:

```
[1] "Loading region london for msoa commute took 0.132 s"
[1] "Loading region london for msoa commute took 0.056 s"
[1] "Plotting top 30 none took 0.01 s"
```

Results:

Peek 2019-09-04 10-45

Robinlovelace:
One question @nikolai-b, when I load the clickable route network I get this output:

```
[1] "Plotting top 30 straight_lines took 3.021 s"
[1] "Plotting top 30 route_network took 0.876 s"
```

Is that right?

nikolai-b:
> One question @nikolai-b, when I load the clickable route network I get this output:
> 
> ```
> [1] "Plotting top 30 straight_lines took 3.021 s"
> [1] "Plotting top 30 route_network took 0.876 s"
> ```
> 
> Is that right?

As in the route network is much faster?  I'm not sure what your asking.

route

Robinlovelace:
I meant are the statements correct. I don't think they are. When you load the route network I think it's incorrect that shiny is `Plotting top 30 route_network took 0.876 s`. Also, I was surprised to see the route network was that much faster. Looking at the route network timings in your gif make more sense.

PR #789, opened 2019-08-30
Test small datasets

AnnaGoodman1:


Robinlovelace:
Great work @AnnaGoodman1, I plan to test this next week, Tuesday most likely. If you could give it a try @nikolai-b or @usr110 before then that would be great. Thanks.

nikolai-b:
So currently London MSOA commute with lines or routes takes (on my laptop) ~ 8.3 s to load the Rds files and an additional ~ 2.3 s to order and plot the lines.  Sadly we currently are doing both of these twice :cry: 

With this MR loading the Rds files goes down to ~ 4.5 s and ordering to ~ 1.7 s.

I needed to make one small change to the data as the file extension needs to be Rds for it to load: https://github.com/npct/pct-outputs-regional-Rsmall/commit/dae58c795199df32d2f5541edc1be1c0667b5dcf but otherwise this all looks good.

scripts/publish-pct-shiny
```diff
@@ -10,7 +10,7 @@ VAR_DIR="/home/git/var-shiny/"
 WAIT_SECONDS=0
 BRANCH=$1
 SHINY_APP_DIR="pct-shiny"
-DATA_GIT="pct-outputs-regional-R"
+DATA_GIT="pct-outputs-regional-Rsmall"
```
nikolai-b: This will expect the `pct-outputs-regional-Rsmall` git repo to already exist on the server so will fail.  I don't think we will change the data repo much so I don't intent to automate this but if we do want to merge it we will have to run the commands in the `Server.md` file manually first.

nikolai-b: This will expect the `pct-outputs-regional-Rsmall` git repo to already exist on the server so will fail.  I don't think we will change the data repo much so I don't intent to automate this but if we do want to merge it we will have to run the commands in the `Server.md` file manually first.

AnnaGoodman1:
Great thanks Nikolai. Any way we can also stop the loading everything twice thing?

AnnaGoodman1:
Probably not easily, and I don't want to suggest code overhaul.... but just checking....

Robinlovelace:
Just picking up on this. Thanks for testing it @nikolai-b - I didn't get time yesterday as initially planned. I think the priority should be to test and merge #788 and then, if that works with these new smaller datasets, merge this one after. So will test #788 now.

nikolai-b:
I've added `pct-outputs-regional-Rsmall` to `npct0` and `npt1`

nikolai-b:
> Probably not easily, and I don't want to suggest code overhaul.... but just checking....

Yes 17350b3809ca24a4c95ebee1dd28259ed4501160 and d8b339e69561e56a6f96387edb45f6ebc5c4f353.  I did them in #788.  The way shiny works it isn't obvious what triggers an observable so it is easy to trigger them when we don't want it (hence the double loads or sorts).

PR #798, opened 2019-09-15
Fix lines are again redrawn

### LINKED ISSUE #797: Lines do not change when moving map
17350b3 stopped the lines being redrawn when the map moves.

nikolai-b:
Sadly this re-introduce double sorting when changing a region.
Fixes #797

### DIFF
--- a/regions_www/m/server.R
+++ b/regions_www/m/server.R
@@ -574,29 +574,34 @@ shinyServer(function(input, output, session) {
     input$scenario
     input$line_order
 
-
-    shinyjs::showElement(id = "loading")
-
-    line_type <- ifelse(input$line_type == 'routes', "routes_quieter", input$line_type)
-
-    local_lines <- sort_lines(region$plot[[line_type]], input$line_type, input$nos_lines)
-    # Filter out zero lines for scenario in question from route network
-    if (input$line_type == "route_network") {
-      if (input$scenario == 'olc') {
-        local_lines <- local_lines[local_lines$bicycle>0,]
-      } else if (input$scenario == 'govtarget') {
-        local_lines <- local_lines[local_lines$govtarget_slc>0,]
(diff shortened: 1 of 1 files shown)


Robinlovelace:
Confirmed issue demonstrated below on master.

Peek 2019-09-15 13-20

Robinlovelace:
Fix demonstrated after merging #798 

Peek 2019-09-15 13-22

Robinlovelace:
Quick check-in question: can you think of any unintended consequences of this @nikolai-b or @usr110 ? I've had a look over the code and quick test and it seems fine but thought it worth checking. Thanks!

nikolai-b:
@Robinlovelace thank you for the quick test but I just wasn't happy that we were re-sorting the lines twice when changing region.  I've fixed this in 440fe026ff1ae352cef44e6dd7b56 and noticed that we were also re-drawing zones af9c840
If your still happy I'll merge.

Robinlovelace:
I think this is a definite improvement although don't have a very strong understanding of the this part of the code. 

@nikolai-b when you say we're re-drawing the zones, could that also be slowing down the initial zones? If so I'd suggest merging this into `master` and `production` and then working on that in a separate issue to keep things moving.

usr110:
Great work @nikolai-b 

Just testing it and found an interesting phenomena. For `london`, specifically, if you move about the map - while the top lines remain as they are, they are redrawn. A lot of changes have been recently made, so I am not sure if it's a known issue or not. Just for comparison, I didn't find the same issue on the live-server.

usr110:
The lines are not redrawn - sorry about it @nikolai-b 

Two things confused me:

1) The plotting message saying that they are redrawn. I know they're used for local testing, but with messages like `Plotting top 30 straight_lines took` for `straight_lines` is slightly confusing.
2) The loading circle - which is a minor issue.

thanks!

PR #805, opened 2019-10-01
Stability

nikolai-b:
I broke object removal in 94bb365, this fixes that and also I noticed systemd can't restart shiny-server as it has forked so we should use `exec`.

### DIFF
--- a/regions_www/m/server.R
+++ b/regions_www/m/server.R
@@ -234,9 +234,11 @@ shinyServer(function(input, output, session) {
     # Rm objects (by time last accessed) if the list size is more than 2 Gb
     if (format(object.size(loaded_data), units = "Gb") > 2) {
       idx_to_rm <- which(loaded_data_accessed == min(unlist(loaded_data_accessed)))
-      print(c("removing", dir_to_rm, format(object.size(loaded_data), units = "Gb")))
-      loaded_data[[idxes_to_rm]] <<- NULL
-      loaded_data_accessed[[idxes_to_rm]] <<- NULL
+      if (interactive()){
+        print(c("removing", names(loaded_data)[[idx_to_rm]], format(object.size(loaded_data), units = "Gb")))
+      }
+      loaded_data[[idx_to_rm]] <<- NULL
+      loaded_data_accessed[[idx_to_rm]] <<- NULL
     }
 
     if (file.exists(filepath)) {
--- a/scripts/shiny-server.service
+++ b/scripts/shiny-server.service
@@ -3,7 +3,7 @@ Description=ShinyServerSHINYN
 
 [Service]
 Type=simple
-ExecStart=/usr/bin/env bash -c '/opt/shiny-server/bin/shiny-server /etc/shiny-server/shiny-server-SHINYN.conf >> /var/log/shiny-server-SHINYN.log 2>&1'
+ExecStart=/usr/bin/env bash -c 'exec /opt/shiny-server/bin/shiny-server /etc/shiny-server/shiny-server-SHINYN.conf >> /var/log/shiny-server-SHINYN.log 2>&1'
 KillMode=process
 ExecReload=/usr/bin/env kill -HUP $MAINPID
 ExecStopPost=/usr/bin/env sleep 5


regions_www/m/server.R
```diff
@@ -234,9 +234,11 @@ shinyServer(function(input, output, session) {
     # Rm objects (by time last accessed) if the list size is more than 2 Gb
     if (format(object.size(loaded_data), units = "Gb") > 2) {
       idx_to_rm <- which(loaded_data_accessed == min(unlist(loaded_data_accessed)))
-      print(c("removing", dir_to_rm, format(object.size(loaded_data), units = "Gb")))
-      loaded_data[[idxes_to_rm]] <<- NULL
-      loaded_data_accessed[[idxes_to_rm]] <<- NULL
+      if (interactive()){
```
Robinlovelace: What's the downside of printing results when not running interactively? I just checked this by adding

```
  if(interactive())  print("hello")
```

in the `shiny::runExample("01_hello")` example and it does indeed print. Happy with this, just curious as to why you've made printing conditional.

scripts/shiny-server.service
```diff
@@ -3,7 +3,7 @@ Description=ShinyServerSHINYN
 
 [Service]
 Type=simple
-ExecStart=/usr/bin/env bash -c '/opt/shiny-server/bin/shiny-server /etc/shiny-server/shiny-server-SHINYN.conf >> /var/log/shiny-server-SHINYN.log 2>&1'
+ExecStart=/usr/bin/env bash -c 'exec /opt/shiny-server/bin/shiny-server /etc/shiny-server/shiny-server-SHINYN.conf >> /var/log/shiny-server-SHINYN.log 2>&1'
```
Robinlovelace: Not 100% sure what this does, but I gather it makes it easier (possible) to restart shiny-server, right?

regions_www/m/server.R
nikolai-b: Just I was nervous about over logging. When run via shiny server these print statements don't appear in the log and I don't think they'd be useful in the server logs.

scripts/shiny-server.service
nikolai-b: Yup.

regions_www/m/server.R
Robinlovelace: Understood, but do server logs contain outputs from `print()` commands? I tested it in interactive and non-interactive settings and cannot see the the print messages when calling Shiny from a non interactive environment. To test you can try this from R:

```r
# no message
shiny::runGitHub(repo = "rstudio/shiny", subdir = "inst/examples/01_hello")
# message from conditional and non conditional print
shiny::runGitHub(repo = "robinlovelace/shiny", subdir = "inst/examples/01_hello", ref = "patch-1")
```

And this from bash:

```r
 # prints message
Rscript -e 'print("hello")
# no message:
Rscript -e 'shiny::runGitHub(repo = "robinlovelace/shiny", subdir = "inst/examples/01_hello", ref = "patch-1")' 
```

nikolai-b: Thanks @Robinlovelace, good point I'm not sure exactly if shiny puts `print` statements into the log.  With your example I wouldn't expect to see the `print` statement in the console, I'd expect it in the log file.  Even shiny doesn't put `print` statements into the logs we don't need these in production.  Does that answer you question about why I made them conditional?  I added them for local debugging only so I didn't want the server to waste resources (albeit tiny) on running them.

nikolai-b:
@Robinlovelace as this has some stability fixes it'd be good for it to get merged in sooner rather than later, I'd prefer we didn't leave production with known bugs for more than 12 days!

regions_www/m/server.R
Robinlovelace: I vote remove the conditionality but cannot see a downside.

Robinlovelace:
>  I'd prefer we didn't leave production with known bugs for more than 12 days!

agreed but it was in your court. Trying to get a balance between speed of changes onto production and ensuring that the review process works - that requires some level of question and answer. Will aim to nudge after a few days next time.

nikolai-b:
@Robinlovelace yes, sorry I'd dropped the ball on this. I meant to say we want the stability improvements, not to imply it was waiting on you.  Asking questions is very legitimate!

PR #822, opened 2020-01-23
Replace cdn with raw urls for #821

### LINKED ISSUE #821: RawGit has reached the end of its useful life
https://cdn.rawgit.com is no more.  It suggests we move to another service like GitHub pages (but that is limited to 1 GB files so is probably not appropriate).  [S3](https://aws.amazon.com/s3/) is probably a very sensible place to store files (and would mean we could have revisions) but that would be quite a bit of work.

The easiest thing would be to use https://www.jsdelivr.com/rawgit (hopefully they won't shut down, looks like it has sponsors but might need to check it more) as we can simply change all the URL and everything should _Just Work_ ™

Robinlovelace:
This fixes #821 for now. Good enough for now, safe enough to try I reckon.


nikolai-b:
If github pages doesn't work then we can use raw git links but I'd prefer to try and avoid it because it is slower and as the content header is wrong it might cause users problems.

Robinlovelace:
I think we should just use this, GH pages is fidly.

Robinlovelace:
Where is the gh-pages branch?

Robinlovelace:
Regarding the point:

> The problem with this is that it takes longer and will send the wrong Content-Header.

A fraction of a second makes little difference. I was planning to announce the release of the schools layer today but that will have to be postponed until next week. Are you happy with a deadline of Wednesday to test alternative approaches and if they're not read accept this @nikolai-b ?

AnnaGoodman1:
I think if @nikolai-b can do it better then great, ok to wait a bit. Thanks both!

PR #834, opened 2020-04-01
HTML tidy

nikolai-b:
I noticed a few of our HTML pages do not have valid HTML :cry: 

I used https://validator.w3.org

### DIFF
--- a/non_www/tabs/commute/download_national.html
+++ b/non_www/tabs/commute/download_national.html
@@ -13,7 +13,7 @@
       </div>
 
       <div class="col-md-12 text-left" style="padding-top:10px;">
-        <h4><a name="natcommutezones"></a>Commuting data for zones</h4>
+        <h4><a id="natcommutezones"></a>Commuting data for zones</h4>
         <p>Download the Census 2011 and scenario attribute data for all zones in England and Wales.</p>
         <ul>
           <li>Zones (local authorities)*:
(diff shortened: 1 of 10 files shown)


Robinlovelace:
The changes look OK to me, great work @nikolai-b. Are they valid now / how can I check on a local version of the branch?

nikolai-b:
Great question @Robinlovelace , sadly we have no ways of checking the HTML hence getting to that state.  I'd be happy to prioritise looking at this as an issue but I doubt it is something easy to fix.  Most browsers would not complain but using deprecated or invalid syntax could cause some browsers to not be able to view the site. Would you like me to add an issue to Github?

> Are they valid now

It does not make our HTML all valid but it removes some errors.  Might be best to review each commit change separately as then you can see why I've done them, it is just 3 small isolated improvements.

Robinlovelace:
> It does not make our HTML all valid but it removes some errors. Might be best to review each commit change separately as then you can see why I've done them, it is just 3 small isolated improvements.

OK I'll checkout this branch and review as best I can on Firefox. Not my area of expertise but they look like minor changes with clear benefit and no downside so should be fine.

Robinlovelace:
National downloads look fine: 

image

Robinlovelace:
Manual and other html files also look fine in Firefox on this branch.

PR #831, opened 2020-03-31
Add link to training materials

Robinlovelace:

PR #835, opened 2020-04-03
Robinlovelace patch 1

Robinlovelace:
From approved PR https://github.com/npct/pct-shiny/pull/831


Robinlovelace:
> could it be moved to under manuals + case studies, before other reports? For most users, the former two more relevant - and less scary...

Good suggestion Anna makes sense to put advance content after case studies.

Robinlovelace:
Heads-up @AnnaGoodman1 I've updated the link by putting the material in Manuals and Training. Was going to create a separate Training section but realised that this is the only training link on there currently. See below:

image

Robinlovelace:
Double checked after fixing merge conflict, looks good.

Robinlovelace:
Replacing #831

Issue #833, opened 2020-04-01
New links to key resources on website


Robinlovelace:
- [ ] Link to training resources in Manual, e.g.: https://github.com/npct/pct-shiny/pull/831
- [ ] Link to essentials document from About page 
- [ ] Link to codebooks from About page 
- [ ] Link to codebooks from Manual

We don't need to add all of them but though it worth documenting, I've tried to put the links in descending order of importance, there may be more links worth adding.

Issue #785, opened 2019-08-05
data downloads from landing page


JDWoodcock:
could we have the data downloads available on the landing page?

Robinlovelace:
Sounds like a good plan to me, many people still don't know they are available.

usr110:
@AnnaGoodman1 could you tell me where the rmd documents for national level datasets for both layers exist? 
For the main page, I think we need to merge both layers in a single document

AnnaGoodman1:
should we maybe just link to national datasets from main page? can't see how linking to all regional ones would easily work, though happy to see ideas

AnnaGoodman1:
rmd docs at: pct-shiny\non_www\tabs\commute and pct-shiny\non_www\tabs\school

usr110:
> should we maybe just link to national datasets from main page

I think so. Regional data would be quite complex.


> rmd docs at: pct-shiny\non_www\tabs\commute and pct-shiny\non_www\tabs\school

Thanks. I don't see any rmd docs there. Have they been moved somewhere else?

See for instance here: https://github.com/npct/pct-shiny/tree/master/non_www/tabs/commute

nikolai-b:
We could improve the regional pop ups on the front page and put regional download links in them.

AnnaGoodman1:
@usr110 , you are right - but those are the download page files I think. I guess html not .rmd...?

usr110:
> 
> 
> @usr110 , you are right - but those are the download page files I think. I guess html not .rmd...?

Thanks. I will work with html then.

Robinlovelace:
Also just picked up on this. Sounds like you're onto a solution @usr110, let us know if you'd like to test/review any changes. Will make the data that bit more accessible, amazing how few people know the data downloads are there.

Robinlovelace:
> What about instead having a link to the regional data download page in the pop-up box, as shown in this image

I'm happy with that it sounds like it could be tricking and time-consuming to implement. My understanding is that @usr110 would work on it and for me it's a question of how long it would take. Previous comments from Ali, e.g. that it 

> will increase complexity

and

> might not work smoothly

led me to proposed a simpler stop-gap solution that points people to the data. I think the file structure is fairly self explanatory and that advertising national and regional data has +s that outweigh potential -s of people being confused when they are directed to GitHub. Are there additional -s of my suggestion, of simply adding a sentence on the main page directing people to the downloads, that we've not considered? 

[...]

PR #836, opened 2020-04-03
Update production

Robinlovelace:
I suggest we merge this after

- [ ] #835 
- [ ] #833
- [ ] #785 

Any others?


nikolai-b:
@Robinlovelace I suggest we merge it now.  Small incremental releases are easier to debug.  That said the other changes you are suggesting are also small so in this case it doesn't matter as much.  Still I'd advocate we [release early, release often](https://en.wikipedia.org/wiki/Release_early,_release_often), it is more agile.

Robinlovelace:
Sounds good @nikolai-b, some of those are small and almost done  - just pushed to https://github.com/npct/pct-shiny/pull/835 so will see if we can get these multiple small changes in by the end of the week to minimise pushes to master, although definitely agree with the principle of lots of small incremental changes being better than deploying everything in one go. All minor, little urgency, low risk so go well as one I think.

Robinlovelace:
The changes listed above are still worth doing. I'll open a separate issue on that.

Summary of changes: 

- Addition of the PCT Essentials document.
- Updated links in the Manual tab

The PDF may well change soon but merging now on the basis of an Essentials document link is infinitely better than none. Sorry it's taken so long to merge, was hoping to get other changes in there.

Discussed with @nikolai-b just now.

PR #850, opened 2020-09-21
Tidy up accessibility HTML

nikolai-b:
Extract the statement into body so we can reuse it in the app.
Use standard jQuery and Bootstrap libraries to keep our site consistent.
Add a link in the nav.


regions_www/nav.html
```diff
@@ -15,6 +15,9 @@
       <li style="margin-left:10px;">
         <a href="/tabs/about.html">About</a>
       </li>
+      <li style="margin-left:10px;">
+        <a href="/tabs/accessibility.html">Accessibility Statement</a>
```
Robinlovelace, line 19: That simple? Simple + effective!

Robinlovelace (approved):

PR #851, opened 2020-09-22
Jekyll

### LINKED ISSUE #848: Accessibility statement on production
Superseding this: https://github.com/npct/pct-shiny/pull/847 

> This is a starter for 10. I think before merging it should have:

* [x]  Link from main page
 
* [x]  css?
 
* [x]  anything else?

nikolai-b:
Fixes #848 

Moving to using a Jekyll.  I can incoporate it into the github push workflow but for now I've just done it manually with:

```sh
cd jekyll-static
jekyll s -d ../regions_www
```

@Robinlovelace any thoughts?


Robinlovelace (changes requested):
This PR tackles two issues simultaneously: the need to use Jekyll (I can see the benefits of static site generators but not sure what the benefits would be vs costs of additional complexity for this fairly simple use case and how it would compare with other options such as Hugo) and the need to update the website. I suggest splitting this PR, that adds over 1000 lines and removes 113, into two 'atomic' PRs unless there is a good reason for bundling the two things together. Thinking: that will make it easier to review, with the update to location of links to the accessibility statement being the priority.

Robinlovelace (approved):

PR #860, opened 2020-10-21
Fix Figure 3

AnnaGoodman1:


usr110:
Hi @AnnaGoodman1 

Thanks for fixing it so quickly. Do we know what really caused it?

Robinlovelace (approved):

PR #867, opened 2020-12-03
Update production

Robinlovelace:


nikolai-b (approved):
I've added a commit updating the Jekyll code.  Without it the new link was missing from http://npct0.vs.mythic-beasts.com/manual.html#casestudies

PR #882, opened 2022-03-15
Update about_body.html

usr110:
Update links for personal page of mine and @JDWoodcock


Robinlovelace (approved):

Robinlovelace:
Good tweaks, worth updating the other links with http:// not https:// URLs?

Robinlovelace (commented):
We could use this as an opportunity to update all URLs. No other ones are broken but good to do it in one go, e.g. for Anna's: https://github.com/npct/pct-shiny/pull/882/files#diff-0b70f2be0647d854c713fc5e284c2d845907c22fa6d5318223d260acd2414090L51

usr110:
> Good tweaks, worth updating the other links with http:// not https:// URLs?

Thanks for reviewing. I don't mind either way. We just need to make sure that none of the links uses HTTPS then. I can see that the blog entry, for instance uses it.

usr110:
> We could use this as an opportunity to update all URLs. No other ones are broken but good to do it in one go, e.g. for Anna's: https://github.com/npct/pct-shiny/pull/882/files#diff-0b70f2be0647d854c713fc5e284c2d845907c22fa6d5318223d260acd2414090L51

Indeed. Not following what you said by Anna's link though. It seems to work fine at my end.

mvl22:
> We just need to make sure that none of the links uses HTTPS then. I can see that the blog entry, for instance uses it.

If a site works under HTTPS, these days you should be linking to that instead of the HTTP equivalent. Very little of the web is now on HTTP, because it produces a browser warning.

Robinlovelace:
Let's update all of them then :+1:

Robinlovelace (changes requested):
Should be https throughout.

usr110:
> if a site works under HTTPS, these days you should be linking to that instead of the HTTP equivalent. Very little of the web is now on HTTP, because it produces a browser warning.

Good point 👍

nikolai-b:
:wave: the checks are failing as the github workflow is broken.  I didn't fully understand the docs and thought I needed to have the workflow in `master` before it would run but I was wrong and I've fixed it in #883 so you can ignore the build error here, sorry.

usr110:
> 👋 the checks are failing as the github workflow is broken. I didn't fully understand the docs and thought I needed to have the workflow in `master` before it would run but I was wrong and I've fixed it in #883 so you can ignore the build error here, sorry.

Got it. Many thanks 👍

Robinlovelace (approved):

Robinlovelace:
Good job on the updates Ali, the failing tests have nothing to do with this cosmetic and good PR so merging.

